PR #55, opened 2021-05-28
refactor load_cmipng_tas and load_cmipng_hfds

mathause:
This adds a common function for `load_cmipng_tas` and `load_cmipng_hfds` as suggested by your comment. The only difference was the use of `np.average` with a masked vs not-masked array. As tas has data everywhere this should not make a difference.

 Next up are then tests and then an internal refactor.

### DIFF
--- a/mesmer/io/_load_cmipng.py
+++ b/mesmer/io/_load_cmipng.py
@@ -519,7 +519,7 @@ def _load_cmipng_var(esm, scen, cfg, varn):
     time, lon, lat, wgt3d = extract_time_lon_lat_wgt3d(data)
 
     # compute area-weighted mean
-    # ATTENTION: does not account for land fraction within grid cells. I.e., coastal
+    # ATTENTION: does not account for land fraction within grid cells. i.e., coastal
     # grid cells count as full ocean grid cells. Expected to have negligible impact on
     # global mean.
 
--- a/tests/test_load_cmipng.py
+++ b/tests/test_load_cmipng.py
@@ -73,8 +73,8 @@ def test_load_cmipng(test_data_root_dir, varn):
     npt.assert_allclose(time, time_expected)
 
 
-@pytest.mark.parametrize("varn", ["tas", "hfds"])  #
-@pytest.mark.parametrize("start, end", [("1850", "1900"), ("1850", "1860")])  #
+@pytest.mark.parametrize("varn", ["tas", "hfds"])
+@pytest.mark.parametrize("start, end", [("1850", "1900"), ("1850", "1860")])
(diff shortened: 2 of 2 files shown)


mesmer/io/_load_cmipng.py
```diff
@@ -514,86 +430,97 @@ def load_cmipng_tas(esm, scen, cfg):
...
+            dta_ref = (
+                data[varn]
+                .sel(time=slice(ref["start"], ref["end"]))
                 .mean(dim="time")
                 .values
             )
-            tas[run] -= tas_ref  # compute anomalies
+            dta[run] -= dta_ref  # compute anomalies
 
         if ref["type"] == "first" and run == "1":
```
mathause, line 493:
```suggestion
        # TODO: remove "first" (used for Leas first paper but does not really make sense)
        if ref["type"] == "first" and run == "1":
```

tests/test_load_cmipng.py
```diff
@@ -0,0 +1,121 @@
...
+    assert dta[1].shape == (251, 20, 20)
+    assert dta_global[1].shape == (251,)
+
+    npt.assert_allclose(lon["c"], lon_expected)
+    npt.assert_allclose(lat["c"], lat_expected)
+    npt.assert_allclose(time, time_expected)
+
+
+@pytest.mark.parametrize("varn", ["tas", "hfds"])  #
+@pytest.mark.parametrize("start, end", [("1850", "1900"), ("1850", "1860")])  #
```
mathause, line 77:
```suggestion
@pytest.mark.parametrize("varn", ["tas", "hfds"])
@pytest.mark.parametrize("start, end", [("1850", "1900"), ("1850", "1860")])
```

mesmer/io/_load_cmipng.py
```diff
@@ -524,86 +440,98 @@ def load_cmipng_tas(esm, scen, cfg):
...
-            tas[run] -= tas_ref  # compute anomalies
+            dta[run] -= dta_ref  # compute anomalies
 
     # extract time, longitude, latitude, and 3d weights from data
     time, lon, lat, wgt3d = extract_time_lon_lat_wgt3d(data)
 
-    # compute area-weighted GSAT from global tas
-    GSAT = {}
+    # compute area-weighted mean
+    # ATTENTION: does not account for land fraction within grid cells. I.e., coastal
```
mathause, line 522:
```suggestion
    # ATTENTION: does not account for land fraction within grid cells. i.e., coastal
```

mathause:
This would be ready for review. More refactoring could be done but let's do this in another PR.

PR #59, opened 2021-06-15
load_phi_gc end to end test

mathause:
a short test for `load_phi_gc`. I only use 2 x 2 grid points - the expected output becomes very big very fast. Once this is in I can then refactor `gaspari_cohn` and `load_phi_gc`.


mathause:
@leabeusch you could have a look at this PR for an example of a smaler test, i.e. the way you would usually design them without any saved files

PR #70, opened 2021-06-24
Remove ens_type keys from all scripts

### LINKED ISSUE #63: Remove ens_type_tr / ens_type from config file and the training scripts
As far as I remember, I only use this as a key in the calibrated parameters dictionaries and for the filename creation of the saved output (calibrated parameters and emulations). I originally introduced it to give additional information about the type of input data that was used for training (i.e., as metadata). But it’s neither intuitive nor does it actually add value -> it should be removed.

leabeusch:
Removes the ensemble type (ens_type) key from all files and thus closes #63.


mathause:
Looks good. Is that not saved in the mesmer bundle? (or why do the tests pass?)

leabeusch:
True, that is strange. @znicholls do you have an idea why this is the case? Shouldn't `test_calibrate_mesmer()` theoretically fail since the old params dictionaries contain 1 more key (`ens_type`) than the new ones?

znicholls:
My bad, fix in #71

leabeusch:
Ok, after rebasing to master, I force pushed my local branch to here and the mesmer calibration test failed (as it should). In a next step, I updated the expected output of `test_calibrate_mesmer.py` (i.e., `test-mesmer-bundle.pkl)` and included the updated test data file into this PR. Now the tests pass again. Please let me know in case this is not the normal / "appropriate" approach to carry out a PR which leads to failing tests.

znicholls:
All perfect nicely done

znicholls (approved):
lgtm

Issue #72, opened 2021-06-27
Add tests checking smaller chunks of the code


leabeusch:
Not too urgent, but definitely something we should work on eventually.

mathause:
I think we got far here - so I am closing this.

PR #73, opened 2021-06-27
Remove full_model key

### LINKED ISSUE #64: Remove the ‘full_model’ key from the params_lt dictionary
This key contains all the individual sklearn models, i.e., 1 sklearn model per grid point (as individual keys). This hinders saving `params_lt` in netcdf format. The loss of information from no longer saving the full sklearn models won’t be problematic since emulations are created based off of the extracted coefficients of the linear regression anyways. On the other hand, the gain should be large as this opens the door to saving `params_lt` in netcdf instead of in .pkl format.

Related to #50 and #18

leabeusch:
Removes the full_model key from params_lt and thus closes #64.

A few points:
- I’m not sure I took the right approach for creating this branch… I just checked out a new branch on top of the `remove-ens_type-key` branch (#70), since I want this one to be merged into `master` after `remove-ens_type-key` is (which is just waiting for approval now)…. But I’m a bit worried this will turn messy because we’ll squash & merge `remove-ens_type-key` into `master` and thus the master history won’t be the same as the one I started this branch from? Should I have instead started this branch by checking out a new branch directly from master?
- When working on this PR I realized, I could really profit from tests that check smaller parts of the code. I.e., since I had to do quite some re-writing within `train_lt()`, I was a bit worried about introducing errors in the parameter calibrations (I double-checked manually on the side). I’m not sure the way we test now, we could really pick those up? I.e., I have to update the whole test output file anyways because a key is missing in the new one -> with that I also introduce whatever new calibration parameters I have obtained. Or can your MESMER calibration test return several assertion errors at once @znicholls? Obviously we’re all already aware that more tests about smaller parts of the code would be good, but I opened #72 for it now nevertheless.
- Same worries about my approach to updating the test files as in #70 -> would be great to hear from you if it’s ok like I started doing it or if this should be done differently.


znicholls:
> * Should I have instead started this branch by checking out a new branch directly from master?

I don't think it matters. Once #70 is merged, we can just rebase this and it should all just work (but let's try tomorrow).

> * When working on this PR I realized, I could really profit from tests that check smaller parts of the code.

If this is your concern then I think we should split this PR in 2: 1) add more tests part 2) make the changes. It's a bit more work but gives much more certainty that we haven't made some bad error.

mesmer/calibrate_mesmer/train_lt.py
```diff
@@ -158,15 +127,50 @@ def train_lt(preds, targs, esm, cfg, save_params=True):
...
-    params_lt, params_lv = extract_additonal_params_lt(params_lt, params_lv)
+            for targ_idx, targ in enumerate(params_lt["targs"]):
+                params_lt["intercept"][targ][gp] = reg.intercept_[targ_idx]
+
+                coef_idx = 0  # coefficient index
+                for pred in params_lt["preds"]:
+                    params_lt["coef_" + pred][targ][gp] = reg.coef_[targ_idx, coef_idx]
+                    coef_idx += 1
+
+                if len(preds_lv) > 0:  # assumption: coefs of lv are behind coefs of lt
```
znicholls, line 168: can this assumption be wrong/checked?

leabeusch: As discussed: added a TODO entry to the `train_lt()` script + opened an issue for it (#74)

leabeusch:
When adding the `TODO` comment to the docstring of `train_lt()`, I realized that I don't actually know where I'm supposed to do the line breaks within the docstring (running `black` doesn't seem to adress this). I did it manually now with an approximate guesswork but I'm pretty sure there must be easy ways to actually have the editor or sth else tell you where to put them / put them automatically, right? :sweat_smile: Suggestions @mathause or @znicholls?

znicholls:
I've actually never figured it out properly. I just use an editor which wraps lines and then put the line breaks in manually where they seem sensible. There are probably automated ways though...

znicholls (approved):

leabeusch:
Thanks for the input @znicholls! I've asked @mathause now too and he suggested I could use the atom editor with the vertical line set to 88 characters & go manually from there. Good to know that even you guys are still doing some things manually. ;)

znicholls:
> Thanks for the input @znicholls! I've asked @mathause now too and he suggested I could use the atom editor with the vertical line set to 88 characters & go manually from there. Good to know that even you guys are still doing some things manually. ;)

Haha very good (I use sublime). I think docstrings are hard because all sorts of things can go in there so auto-formatters quickly explode

Issue #75, opened 2021-06-30
Introduce alternative sample weighting approaches into the AR process fitting


leabeusch:
Currently the AR processes in `train_gv()` and `train_lv()` are fit under the assumption that each available scenario has the same weight, regardless of the amounts of initial condition ensemble members it contains. The most straightforward alternative would be to give the same weight to each sample (meaning that scenarios with more initial condition ensemble members would end up with higher weights overall), which is already available for the `train_lt()` estimation. Other extensions could be considered too.

PR #76, opened 2021-06-30
Improve AR processes aspects

leabeusch:
This PR finally updates the methodology in `train_gv()` to follow the same principle as the one in `train_lv()`. I.e., instead of concatenating all runs and estimating the AR parameters from a single time series, the AR parameters are estimated on every individual run and subsequently averaged (first across runs then across scenarios). Additionally, a check is added to both `train_gv()` and `train_lv()` to ensure that an error is raised in case a different approach than equal weight for each scenario is chosen (-> I opened issue #75 as a reminder to eventually add alternative weighting approaches). Lastly, the ability to draw realizations from AR(0) processes is added and docstrings / comments are updated.
Again, I could have profited from #72, but since it would take me quite some effort to learn how to write a proper test, I instead just manually checked that a) the results of the new approach are plausible, b) the output for the old and the new approach is identical if a single run from a single scenario is given. I’ll invest time into learning this type of testing once my manuscript is submitted. This PR will require updating both test output files, as we’re changing the estimated parameters and thus also the drawn emulations.
Open question: is it ok to use the `raise ValueError` feature as I do here, or should another error be raised instead? This error gets raised if `wgt_scen_tr_eq` is `False` instead of `True` because currently only an “equal weights for each scenario regardless of number of initial condition ensemble members”-approach is implemented for the AR process parameter estimations.


mesmer/calibrate_mesmer/train_gv.py
```diff
@@ -131,59 +137,83 @@ def train_gv_AR(params_gv, gv):
...
-        k = (
-            gv[scen].shape[0] * gv[scen].shape[1]
-        )  # = nr_runs*nr_ts for this specific scenario
-        gv_all[i : i + k] = gv[scen].flatten()
-        i += k
-
     # specifiy parameters employed for AR process fitting
-    max_lag = 12  # rather arbitrarily chosen in trade off between allowing enough lags and computational time
-    # open to change
+    max_lag = 12  # rather arbitrarily chosen in trade off between allowing enough lags
```
znicholls, line 154: keyword argument?

znicholls: if "max_lag" not in kwargs:
    kwargs["max_lag"] = 12

mesmer/calibrate_mesmer/train_gv.py
```diff
@@ -11,13 +11,13 @@
 
 import joblib
 import numpy as np
-from statsmodels.tsa.ar_model import ar_select_order
+import statsmodels.api as sm
+from statsmodels.tsa.ar_model import AutoReg, ar_select_order
 
 
 def train_gv(gv, targ, esm, cfg, save_params=True):
```
znicholls, line 18:
```suggestion
def train_gv(gv, targ, esm, cfg, save_params=True, **kwargs):
```

mesmer/create_emulations/create_emus_gv.py
```diff
@@ -84,7 +85,7 @@ def create_emus_gv(params_gv, preds_gv, cfg, save_emus=True):
             )
         else:
             raise ValueError("The chosen method is currently not implemented.")
-            # if the emulations should depend on the scenario, scen needs to be passed to the fct
+            # if the emuus should depend on the scen, scen needs to be passed to the fct
```
znicholls, line 88:
```suggestion
            # if the emus should depend on the scen, scen needs to be passed to the fct
```

znicholls (approved):
Small suggestion about how to make things a bit tidier. Otherwise over to you to merge when you're happy or request a new review.

leabeusch:
Can you quickly check this again please @znicholls? Not entirely sure I put everything at the right place. Also: what should a `**kwargs` docstring entry look like?? (needed in `train_gv()`)

mesmer/calibrate_mesmer/train_gv.py
leabeusch: addressed in commit: 44c8f79e8dc634062778bc0db7d83b10428e1ec0

mesmer/calibrate_mesmer/train_gv.py
leabeusch: addressed in commit: 44c8f79e8dc634062778bc0db7d83b10428e1ec0

znicholls:
> what should a `**kwargs` docstring entry look like?? (needed in `train_gv()`)

Search for 'kwargs' in https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

mesmer/calibrate_mesmer/train_gv.py
```diff
@@ -33,6 +33,8 @@ def train_gv(gv, targ, esm, cfg, save_params=True):
         config file containing metadata
     save_params : bool, optional
         determines if parameters are saved or not, default = True
+    **kwargs:
+        additional arguments
```
znicholls, line 37:
```suggestion
        additional arguments, passed through to the training function
```

znicholls (approved):

PR #78, opened 2021-07-06
readthedocs configuration 

### LINKED ISSUE #29: add documentation on readthedocs
- the repo probably needs to be public for this
- currently that would only be the docstrings - need to think about other content
  - installation requirements
  - minimal example (e.g. based on the test)
  - references -> list relevant papers

probably after #28

mathause:
- [x] closes #29

This tries to get the documentation on rtd running. Currently uses pip to install packages. We may eventually need to switch to conda (one we require cartopy), but until then let's use pip which is faster.

The name mesmer was already taken so it's now under https://mesmer-emulator.readthedocs.io/en/stable/ (well once it works...)


mathause:
No we need conda - we have cartopy in our dependency tree.

mathause:
Ok got it to work, you can check out the docs at: https://mesmer-emulator--78.org.readthedocs.build/en/78/

leabeusch (approved):
Wow, so efficient! Thanks for getting this done so quickly @mathause! Just 1 comment: When looking at the RTD page, I realized that for some reason `train_gv` is listed before `train_gt` in the `api.rst` file & thus also on RTD. From the workflow & the alphabet it would make more sense to have `train_gt` before `train_gv` though. Do you have a reason for the current order or can we change it? (If we change the order I'd probably sneak this change into this PR too, but if you prefer we can also open a new PR instead)

mathause:
No I just copied them in in random order. Certainly fine to change them.

mathause:
Ok let's merge this - we can always update the documentation.

mathause:
@znicholls I added you as maintainer on readthedocs. I also added a stable branch. Currently stable is the same as master (because the rtd config is missing in version 0.1.0-beta). We can do it properly after the next release.

@leabeusch the idea of the stable branch is that it points to the version that was last released. Than means we (usually) only touch this branch just after an update.

znicholls:
> @leabeusch the idea of the stable branch is that it points to the version that was last released. Than means we (usually) only touch this branch just after an ~update~ release.

@mathause I thought read the docs did this automatically based on the tags? https://docs.readthedocs.io/en/stable/versions.html

mathause:
Yes, you are right. Didn't know that. I think it actually tried and failed because the current tag is not correctly configured for rtd. in this case I'll remove the stable branch again.

PR #80, opened 2021-07-13
Update example script

leabeusch:


mathause (approved):
Looks good. But I haven't tested it.

leabeusch:
I actually tested it now and noticed that we had still an old pathway to the test data in the config file -> I update this one plus added full global emulation creation in the example script to also highlight this MESMER feature + I made a few other minor changes. Whether the output is reasonable I didn't check. But I think this is fine for now, especially since we intend to create actually nice examples in the near(ish)-term future.

Issue #62, opened 2021-06-15
Use approximation for great circle distance?


mathause:
Currently we use `geopy.distance.distance` to calculate distances between gridpoints. This uses an exact algorithm, assuming an ellipsoid. For an example dataset of 25 * 53 gridpoints it takes 5min 30 sec to calculate all distances. I think we could reduce this to about 2 min.

However, if we assume a sphere and use the Haversine formulat we can get it down to << 1 sec. The error should generally be below 0.5 % (to be confirmed). An additional advantage is that we no longer need to save the distance and ghi_phi matrices.

To be checked and cleaned:
```python
import numpy as np

def distance(s_lat, s_lng, e_lat, e_lng):
 
    # approximate radius of earth in km
    R = 6371.009
 
    s_lat = s_lat*np.pi/180.0
    s_lng = np.deg2rad(s_lng)
    e_lat = np.deg2rad(e_lat)
    e_lng = np.deg2rad(e_lng)
 
    d = np.sin((e_lat - s_lat)/2)**2 + np.cos(s_lat)*np.cos(e_lat) * np.sin((e_lng - s_lng)/2)**2
 
    return 2 * R * np.arcsin(np.sqrt(d))

def dist_all(LON, LAT):
     dist = np.empty((len(LON), len(LAT)))
     for i, (lon, lat) in enumerate(zip(LON, LAT)):
         dist[i, :] = distance(lat, lon, LAT, LON)
     return dist
 
```
Current implementation:

```python
import geopy.distance

def calc_geodist(lon, lat):

# from geographiclib.geodesic import Geodesic
# g = Geodesic(6378.137, 1 / 298.257223563)
# g.Inverse(47, 8, 0.589, 0.2593, Geodesic.DISTANCE)["s12"]

    n_points = len(lon)

    geodist = np.zeros([n_points, n_points])

    # could be sped up by only computing upper or lower half of matrix
    # since only needs to be done 1x for every land threshold, not implemented
    for i in np.arange(n_points):
        for j in np.arange(n_points):
            geodist[i, j] = geopy.distance.distance((lat[i], lon[i]), (lat[j], lon[j])).km

        if i % 200 == 0:
            print("done with gp", i)

    return geodist
```

Example:

```python

import xarray as xr

air = xr.tutorial.open_dataset("air_temperature")

lon, lat = air.lon, air.lat
LON, LAT = xr.meshgrid(lon, lat)

dist_exact = calc_geodist(lon, lat)
[...]

PR #85, opened 2021-07-22
Refactor phi gc

mathause:
- [x] Tests added

After merging #59 I did a small refactoring of the `load_phi_gc` functionality. 

- vectorized `gaspari_cohn`
- Added `calc_geodist_exact` to it's own function
  - Used `geographiclib.geodesic.Geodesic` which is slightly faster than `geopy.distance` (geopy calls geographiclib under the hood)
  - only calculate the upper half of the matrix
- added some unittests for `gaspari_cohn` and `calc_geodist_exact`

TODOs (in another PR):
- This should not live inside `load_constant_files.py`. But I am not sure how to name the file: `distance_utils.py`? `regularization.py`? Something better?
- I would still like to do #62.


CHANGELOG.rst
```diff
@@ -27,6 +27,10 @@ Documentation
 Internal Changes
 ^^^^^^^^^^^^^^^^
 
+ - Refactor and speed up of the gaspari_cohn function and the calculation of the great
```
mathause, line 30:
```suggestion
 - Refactor and speed up of the Gaspari-Cohn function and the calculation of the great
```

mathause:
Anyone planing on reviewing this? With #59 in place this should be a safe change so I will merge soonish if I don't hear back from you.

mesmer/io/load_constant_files.py
```diff
@@ -37,29 +36,62 @@ def gaspari_cohn(r):
...
+        + 5 / 8 * r_s ** 3
+        - 1 / 2 * r_s ** 4
+        + 1 / 12 * r_s ** 5
+        - 2 / (3 * r_s)
+    )
+
+    return y.reshape(shape)
+
+
+def calc_geodist_exact(lon, lat):
```
mathause, line 74:
```suggestion
def calc_geodist_exact(lon, lat):
    """exact great circle distance based on WSG 84

    Parameters
    ----------
    lon : array-like
        1D array of longitudes
    lat : array-like
        1D array of latitudes

    Returns
    -------
    geodist : np.array
        2D array of great circle distances.
    """
```

leabeusch (approved):
Finally had a (too?) quick look at this. Looks all very nice at first glance (also yayy for now actually tests being added as parts of PRs. :thumbsup:). I didn't go into the code in detail (I assume / hope all is fine with it since it doesn't break your `phi_gc` test), but I do have a remark about how to name it: I would not move this to a script called `regularization.py`. The regularization itself happens in `train_lv()` (in Lines 349 & 382). Potentially `regularization_utils.py`? Although I personally wouldn't expect distance calculations in a script like that at all (but instead statistical methods for L1 / L2 regularizations etc).... so maybe just keep the `gaspari_cohn()` & `load_phi_gc()` part in the `regularization_utils.py` script and add yet another script just with the distance calculations called `distance_utils.py`? Then you could also put your great circle approx. there once you have it. Not sure if this is starting to be too many individual files then. But then again, usually you guys do encourage me to separate things more. :smile:

mathause:
Thanks for the feedback. I think I would probably move `calc_geodist_exact` to `mesmer.utils.geodist`. `gaspari_cohn` could go to `mesmer.utils.statistics` (or `mesmer.utils.kernels` (is it a kernel?)). But indeed one function per file is a bit overkill. So both could go to `mesmer.utils.computation`? (Naming is hard!)

PR #90, opened 2021-08-24
Salvage cmip5-ng

### LINKED ISSUE #89: does loading cmip5 data actually work?
In `find_files_cmipng` for cmip5 you only glob `path_runs_list`

https://github.com/MESMER-group/mesmer/blob/04240a4633292f2ed5058a72e3f5292e48ee1e62/mesmer/io/load_cmipng.py#L101

but further down you loop `path_runs_list_scen` which does not exist for cmip5:

https://github.com/MESMER-group/mesmer/blob/04240a4633292f2ed5058a72e3f5292e48ee1e62/mesmer/io/load_cmipng.py#L190

so I think this code path is dead. Do we need to salvage this? @leabeusch

leabeusch:
Fixes issues with reading in data from cmip5-ng archive for all RCPs

 - [ ] Closes #89
 - [ ] Tests added -> not done. Is it needed? Or will that be covered by your PR #55 anyways? If yes, what type of test would you suggest?
- [ ] Introduces `use_cftime=True` to deal with incompatible calendar for default xarray approach (-> thanks to @mathause for suggesting this solution!)
- [ ] Adds ability to deal with fact that *r1i1* member does not always exist (which if exists is always moved to first position due to legacy reasons)
- [ ] Removes tas run with bug in time axis in scenario that wasn't used before

### DIFF
--- a/mesmer/io/load_cmipng.py
+++ b/mesmer/io/load_cmipng.py
@@ -91,8 +91,13 @@ def find_files_cmipng(gen, esm, var, scenario, dir_cmipng):
         dir_name = dir_cmipng + var + "/"
 
         if var == "tas":
-            esms_excl = ["GISS-E2-H"]  # list of all ESMs which have excluded runs
-            runs_excl = ["tas_ann_GISS-E2-H_rcp85_r2i1p1_g025"]  # list of excluded runs
+            esms_excl = ["GISS-E2-H", "EC-EARTH"]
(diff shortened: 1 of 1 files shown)


mathause:
Looks ok to me. The changes should not cause a problem for #55 I think - we don't touch the same functions. Tests would be nice but testing IO is one of the more difficult things to do so I'd leave it as is.

mathause (approved):

PR #93, opened 2021-10-01
Release v0.8.2

leabeusch:

A small PR to have a commit that says v0.8.2


mathause:
The change looks good - I like commits that say "Release x.y" so they can be found in the history. Do you have an idea what the test failure is? It is certainly not related but would be good to fix before releasing. (In another PR).

leabeusch:
> I like commits that say "Release x.y" so they can be found in the history

Is this just a confirmation that this is a good PR or do you want me to change the name from "Release vX.Y" to "Release X.Y"? :sweat_smile: (I just followed our example of PR #81 here)

> Do you have an idea what the test failure is? 

Not at all. I'm completely surprised by this & was hoping you'd know why it occurs. I mean we didn't change anything in any other file than the changelog.rst since the last PR that still passed all the tests. Maybe (hopefully!) @znicholls has an idea what this could be about?

mathause:
> Is this just a confirmation that this is a good PR or do you want me to change the name from "Release vX.Y" to "Release X.Y"?

All good as is.

Concerning the error - it fails to multiply two matrices of differing size, one of which is `"phi_gc"`

```python
                ecov = np.cov(y_est, rowvar=False, aweights=wgt_scen_eq_est)
>               loc_ecov = aux["phi_gc"][L] * ecov
E               ValueError: operands could not be broadcast together with shapes (117,117) (118,118)
```
it does work locally so might be an issue with an update of one of our dependencies. Not sure yet.

mathause:
Ok I could trace the difference back to the update of regionmask - now I trying to find out why the mask changed. :grimacing:

mathause (approved):
Ok let's merge #96, then #94 and then this. In #96 I slightly changed the format of CHANGELOG so we will likely run into merge conflicts here. I can take care of this.

PR #94, opened 2021-10-01
ci: list installed packages

mathause:
This adds a `conda list` step to the ci. I like to have this to be able to compare installed package versions. Also renames one job from "build" to "test" which I think is better.


znicholls (approved):
Nice

mathause:
thanks for the review!

PR #96, opened 2021-10-01
recreate test files after regionmask update

### LINKED ISSUE #95: test failure with regionmask 0.8
#93 surfaced a test failure:

```python
    ecov = np.cov(y_est, rowvar=False, aweights=wgt_scen_eq_est)
>   loc_ecov = aux["phi_gc"][L] * ecov
E   ValueError: operands could not be broadcast together with shapes (117,117) (118,118)
```

one of the arrays changed size.

---

I can trace the issue back to the update of regionmask to v0.8.0: it's due to regionmask/regionmask#213 (fixed in regionmask/regionmask#271) - the good news is that this issue avoided regionmask/regionmask#266 (which would've been even worse) and it's now correct in v0.8.0 :shrug: :thinking:

The issue was that regionmask determined the grid to be from -180...180, while it was -188.1...170.1. 

<details>

```python
import numpy as np
import regionmask
import mesmer

# the test dataset has the following coords:
lon = np.arange(-180, 180, 18)
lat = np.arange(-85.5, 90, 9)

lons = mesmer.utils.regionmaskcompat.sample_coord(lon)
lats = mesmer.utils.regionmaskcompat.sample_coord(lat)

regionmask.core.mask._is_180(lons.min(), lons.max())
```

```python
import cartopy.crs as ccrs
import matplotlib.pyplot as plt

f, axs = plt.subplots(1, 2, subplot_kw=dict(projection=ccrs.PlateCarree()))
region = regionmask.defined_regions.natural_earth.land_110

[region.plot(ax=ax, add_label=False) for ax in axs]
ax = axs[0]
region.mask(lons, lats, wrap_lon=False).fillna(1).plot(ax=ax, add_colorbar=False)
ax.set_title("regionmask v0.6.2")

ax = axs[1]
region.mask(lons, lats).fillna(1).plot(ax=ax, add_colorbar=False)
ax.set_title("regionmask v0.8.0")
```

</details>

See the right side of the two panels for the difference:

different_masks

---

Now I need to find out why only _one_ of the arrays changed it's size (and which is the correct one).

mathause:
- [x] Closes #95 
 - [x]  Tests added


mathause:
Argh I accidentally pushed to master... that is now reverted and here is the PR.

---

I had to reproduce the test files because there was a bug-fix in regionmask v0.8.0 which adds one more grid point in our tests. What I am not sure if we should now require regionmask >= 0.8. I suggest to leave it as is.


@leabeusch how urgent is #93?

mathause:
The tests pass and there is nothing to review, so I am merging.

leabeusch:
Thanks for all your work on this. To answer your question with a few days delay: depends on what you consider urgent. I've submitted a paper last week which states that  v0.8.2 of MESMER was used for training. Hence it'd be great if we managed to have an actual version 0.8.2 available that contains the code with functionality as of ~end of August (i.e., the one I originally tried to have my PR for) at some point in the not too distant future.

PR #97, opened 2021-10-04
no longer save phi_gc file

mathause:

The slow part is the calculation of the geo distance. So let's remove the caching of the phi_gc file. Should be merged after the new release. I also wait with a CHANGELOG entry.


mathause:
I am closing this as out of date. Might come back to this later.

PR #99, opened 2021-10-04
enable codecov

mathause:

This should upload the code coverage to codecov.io. Let's see if this already works or if it needs some changes...


mathause:
@znicholls any objections?

.codecov.yml
```diff
@@ -0,0 +1,11 @@
+codecov:
+  require_ci_to_pass: yes
+
+coverage:
+  status:
+    project:
+      default:
+        # Require 1% coverage, i.e., always succeed
```
znicholls, line 8: Wouldn't be my first choice but fine for now

mathause: I had very strict settings in regionmask (coverage needed to be `>=` for a PR). However, sometimes the coverage goes down because a line of code is deleted which is just annoying. So I am now doing this. But I am of course open to change it.

znicholls:
Nice let’s start here then

PR #108, opened 2021-10-19
#98 follow up: fix changelog

mathause:
In #98 I added the changelog entry to the wrong version...


mathause:
:stuck_out_tongue_closed_eyes:

Issue #106, opened 2021-10-14
internal data structure


mathause:
NOTE: this issue is very much a draft at the moment.

---

We should give some thought on the internal data structure in mesmer. IMHO this is one of the more important but also difficult things to decide on. Generally the idea is to move to a xarray-based data structure. However, it is unclear how we want to carry metadata around (e.g. model name, experiment, etc.).

# Current status

Currently a triply nested dictionary of numpy arrays. Something that roughly looks like:

```python
arr = "array"
data = dict(model=dict(ssp=dict(tas=arr)))
```

# Ideas

## 1. Keep as is

**Pros/ Cons**
- No need to change anything
- Data access is awkward (`data["CESM"]["ssp585"]["tas"]`)
- What if another layer is needed?
- ...

## 2. One huge DataArray

We could potentially create one enormous DataArray that encapsulates all necessary information as coordinates (and non-dimension coordinates). It could look something like this:

```
<xarray.DataArray (time, ens, gp, realisation)>
array([[1, 2, 3]])
Coordinates:
    model    (ens) <U5 'CESM' 'CESM' 'MIROC'
    exp      (ens) <U6 'ssp585' 'ssp126' 'ssp585'
  * time     (time) int32 1
Dimensions without coordinates: ens, gp, realisation
```

**Pros/ Cons**
- A well understood and well tested data structure
- Can get unwieldy if too big
- Difficult to decide where to draw the border
- Need to be careful what becomes a dimension (e.g. not all models have all scenarios, so it cannot be a dimension)
- Cannot have hist and projections in the same array
- Cannot combine different models if they have a different grid

## 3. DataList

_DataList_ is a data structure inspired by [ESMValTool](https://github.com/ESMValGroup/ESMValTool/) that I used extensively for my analysis of the CMIP6 data for IPCC. It is used as a flat data container to store model data (e.g. `xr.DataArray` objects) and its metadata. It is a list of data, metadata tuples:
```python
datalist = [
    (ds, meta),
    (ds, meta),
    ...
]
```
where `ds` is a data store (e.g. an `xr.DataArray`) and meta is a `dict` containing metadata, e.g. `meta = {"model": "CESM", exp: "ssp585", ...}`. This allows to (1) work with a convenient flat data structure (no nested loops), (2) store arbitrary data (e.g. `xr.DataArray` objects with different grids), (3) carry around metadata without having to alter the data store.

[...]

PR #109, opened 2021-10-21
Add prototype of class structure

znicholls:
- [ ] Closes #xxx
 - [ ] Tests added

### DIFF
--- a/tests/integration/test_prototype.py
+++ b/tests/integration/test_prototype.py
@@ -1,5 +1,4 @@
 import numpy as np
-import numpy.testing as npt
 import pytest
 import xarray as xr
 from statsmodels.tsa.arima_process import ArmaProcess
@@ -196,7 +195,7 @@ def test_prototype_train_lt():
         ("global_variability", res_legacy[1]["coef_gvtas"]["tas"]),
         ("intercept", res_legacy[0]["intercept"]["tas"]),
     ):
-        npt.assert_allclose(res_updated.sel(predictor=updated_name), legacy_vals)
+        np.testing.assert_allclose(res_updated.sel(predictor=updated_name), legacy_vals)
 
 
 def _do_legacy_run_train_gv(
@@ -233,9 +232,9 @@ def _do_legacy_run_train_gv(
     ),
 )
 def test_prototype_train_gv(ar):
-    time_history = range(1850, 2014 + 1)
-    time_scenario = range(2015, 2100 + 1)
-    time = list(time_history) + list(time_scenario)
+    time_history = np.arange(1850, 2014 + 1)
+    time_scenario = np.arange(2015, 2100 + 1)
+    time = np.concatenate((time_history, time_scenario))
 
     magnitude = np.array([0.1])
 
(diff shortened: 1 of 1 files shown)


znicholls:
@mathause this is sort of what I was thinking (I didn't get up to actually writing tests so that will have to wait for another day). There's a lot of boilerplate code around the actual calibration. If we can move to some sane classes then it might become much clearer what is actually doing calibration and what is just being copied around because there's not enough utility code available (e.g. this loop https://github.com/MESMER-group/mesmer/blob/d73e8f521a2e1d081a48b775ba14dd764cb671e8/mesmer/calibrate_mesmer/train_gv.py#L174 and this loop https://github.com/MESMER-group/mesmer/blob/d73e8f521a2e1d081a48b775ba14dd764cb671e8/mesmer/calibrate_mesmer/train_lv.py#L239 are the same idea but it's really hard to see at the moment). That will hopefully make it much easier to see how to scale things, how to add new things and where utility code is required instead.

leabeusch:
Based on this comment of @mathause https://github.com/MESMER-group/mesmer/issues/106#issuecomment-953417775, I guess you've moved beyond @znicholls original option proposals but I just want to stress nevertheless, that I'm very against option 3, i.e.,

> calibrate one scenario at a time (doesn't work I don't think because the regressions need everything at once)

as @znicholls already points out himself: this really goes against the general idea of MESMER to be calibrated on a broad range of scenarios simultaneously to ensure that the resulting parameters are able to successfully emulate a broad range of scenarios too i.e., what we analysed in the GMD paper... how good or bad the single scenario approach would be of course always depends on what scenario is used for calibration again & so on... but please don't kill MESMER's overall capability to be trained on multiple scenarios at once in this whole refactoring exercise. :sweat_smile:

znicholls:
Lessons learnt so far from this re-write:

1. Keeping calibration and emulation next to each other would make it much easier to understand what is going on with each part of the model/bit of code
1. Ensuring that methods/functions are no more than 10 lines (with very few exceptions) would make it much easier to see what is going on. Functions with more than 10 lines have too much context, which makes it hard to focus on what is actually going on.
1. Using long, descriptive names also makes a massive difference. Abbreviations make it really hard to keep track of what is going on (and, when combined with long functions, become almost impossible to keep track of).
1. Doing a re-write alongside the original writer would probably be much simpler as they know the theory (I had to do a lot of googling to work out what I was actually looking at)

I have train local variability left to sort out, then I'll close this PR and start with some much smaller steps. I think doing this has given me enough experience to have a sense of how to start with a new structure.

mesmer/prototype/utils.py
```diff
@@ -0,0 +1,136 @@
...
+    dict[float: :obj:`xr.DataArray`]
+        Gaspari-Cohn correlation matrix (values) for each localisation radius (keys)
+
+    Notes
+    -----
+    Values in ``localisation_radii`` should not exceed 10'000 by much because
+    it can lead to ``ValueError: the input matrix must be positive semidefinite``
+    """
+    # I wonder if xarray can apply a function to all pairs of points in arrays
+    # or something
```
mathause, line 36: Good question. I don't think there is a way around a meshgrid. But then we already pass the flattened array.

mesmer/prototype/utils.py
```diff
@@ -0,0 +1,136 @@
...
+    :obj:`xr.DataArray`
+        Gaspari-Cohn correlation function applied to each point in ``inputs``
+    """
+    inputs_abs = abs(inputs)
+    out = np.zeros_like(inputs)
+
+    sel_zero_to_one = (inputs_abs.values >= 0) & (inputs_abs.values < 1)
+    r_s = inputs_abs.values[sel_zero_to_one]
+    out[sel_zero_to_one] = (
+        1 - 5 / 3 * r_s ** 2 + 5 / 8 * r_s ** 3 + 1 / 2 * r_s ** 4 - 1 / 4 * r_s ** 5
```
mathause, line 118: Unrelated but we could speed this up by using
```suggestion
        1 - (5 / 3 + (5 / 8 + (1 / 2 - 1 / 4 * r_s) * r_s) * r_s) * r_s**2
```

(admittedly harder to read but about 10 times faster).

leabeusch:
> Doing a re-write alongside the original writer would probably be much simpler as they know the theory (I had to do a lot of googling to work out what I was actually looking at)

Haven't properly been keeping up with all the refactoring progress that has been going on lately, but in case I could be of help with this point, I assume you'd let me know? ^^'

mesmer/prototype/utils.py
mathause: see #157

Issue #65, opened 2021-06-21
Introduce ability to save output (i.e., calibrated parameters and emulations) to netcdf files


leabeusch:
This is a generalization of #18 and requires #64 to be closed before it can be tackled (#63 closed would be nice too but is not necessary). (#50 is related too)
Probably the current approach of saving to .pkl files should be removed and solely the saving to netcdf option should remain? Probably the new netcdf approach will involve xarray? I’m confident @mathause will know what’s the right way forward here.

mathause:
We also need to think about how we want to lay out the calibrated params on disk. And what is saved as netCDF and what not. Without spending too much time to think about this I would imagine a folder like structure. Maybe something like this: 

```
mesmer_calibrations/
└── 2021_11_03_NAME_OF_STATISTICAL_MODEL
    ├── AWI
    │   ├── meta.yaml
    │   ├── step1_global_variability_AWI.nc
    │   └── step2_local_trend_AWI.nc
    ├── CESM2
    │   ├── meta.yaml
    │   ├── step1_global_variability_CESM2.nc
    │   └── step2_local_trend_CESM2.nc
    └── conf.yaml
```

- A global folder for mesmer_calibrations.
- A folder for each calibration.
- A subfolder for each model (although #13).
- A netCDF for each calibration step.
- `conf.yaml` contains global config values
- `meta.yaml` contains model specific info (could also be attrs in the netCDFs)


`meta.yaml` could look like

```
config_version: 1
mesmer_version: 1.0.0

statistical_model: NAME_OF_STATISTICAL_MODEL

grid_landfraction: 0.33
other_global_config_value: ...


version_info:
- regionmask: 0.8
- xarray: 0.20.0
- ...
```

<details>

and `meta.yaml`
```
model: CESM2

input_data_meta:
- ens: r1i1p1f2
  exp: historical
  model: CESM2
  variable: tasmax
- ens: r1i1p1f1
  exp: ssp119
  model: CESM2
  variable: tasmax
```

</details>

Am I over engineering things?

mathause:
Done, although we somewhat shift the naming problem to the user (although we do save them in tutorials and tests so there is some guidance).

PR #114, opened 2021-11-01
Add ROADMAP.md

### LINKED ISSUE #111: draft roadmap
We (I)  should draft a high level roadmap and possibly a time line...

 @MESMER-group/mesmer

mathause:
- [x] Closes #111


yquilcaille:
> The array data corresponds to climate model data as stored on disk, with dimensions time x longitude x latitude.

Do we move to latitude x longitude, or do we stick to cells, ie gridpoints? I think that if we stick to the emulation of land variables, cells is better for memory reasons.

1. To my perception, the most significant update would be the internal data structure.
2. @leabeusch had already some separation of data handling from the statistical tasks, here it would need to have separate functions for the treatment of loaded & treated data to data ready to use in statistical tasks. For this part, I think that we can either check from a top-down perspective (what are the statistical tasks, what common format of data can they use) to identify what data treatment tasks to perform, or from a bottom-up perspective (what is my starting format of data, what data treatment tasks may be necessary) to identify what final format of data to use in statistical tasks.
3. Persisting calibrated parameters, configuration infrastructure, removing ETH specific code paths may be easier to do.
4. i'm not familiar enough with the code test coverage, i dont know here.

Regarding a timeline, I think that it would mostly depend on the points 1 and then 2.

mathause:
> Do we move to latitude x longitude, or do we stick to cells, ie gridpoints? I think that if we stick to the emulation of land variables, cells is better for memory reasons.

Good point. I think we should continue to use `cells` internally. Optimally MESMER  is able to handle both. It probably does not matter too much for the roadmap.

> 1. To my perception, the most significant update would be the internal data structure.

Yes I agree.

> 2. @leabeusch had already some separation of data handling from the statistical tasks, here it would need to have separate functions for the treatment of loaded & treated data to data ready to use in statistical tasks. For this part, I think that we can either check from a top-down perspective (what are the statistical tasks, what common format of data can they use) to identify what data treatment tasks to perform, or from a bottom-up perspective (what is my starting format of data, what data treatment tasks may be necessary) to identify what final format of data to use in statistical tasks.

True! Let's call it a "better" separation ;-).

> 3. Persisting calibrated parameters, configuration infrastructure, removing ETH specific code paths may be easier to do.

Yes this should be the easier tasks.

ROADMAP.md
```diff
@@ -0,0 +1,76 @@
...
+
+Running MESMER currently involves a mixture of data loading, data preprocessing, defining the emulation model for calibration as well as for drawing realizations. We hope to improve on this by addressing the following points:
+
+- Clearer separation of individual steps from data loading and preprocessing.
+- More intuitive way to define a "emulation model".
+- ...
+
+### Internal data structures
+
+There are two levels of internal data structures used within MESMER: (i) an array data structure, and (ii) a high level structure. The array data corresponds to climate model data as stored on disk, with dimensions `time x longitude x latitude`. The high level data comprises the dimensions: `model x scenario x realization x variable`.
```
leabeusch, line 20: same point as @yquilcaille  about `"cell"` rather than `longitude x latitude` & maybe the point from above again about us having too many `"model"` terms in our vocabulary... potentially `"esm"` here... but that would of course be an abbreviation again which there seems to be some resistance against.

ROADMAP.md
```diff
@@ -0,0 +1,76 @@
...
+
+#### Array data structure
+
+MESMER currently uses [numpy](https://numpy.org/) as array library. which we plan to replace with [xarray](http://xarray.pydata.org/en/stable/). xarray adds named dimensions, coordinates and attributes on top of raw numpy-like arrays and is particularly tailored to working with netCDF files.
+
+Specific tasks
+- Rewrite the internal handling of coordinates and dimensions. Currently time, latitude, longitude etc are passed as individual variables - using xarray this will no longer be necessary.
```
leabeusch, line 27: Probably not relevant for the road map but unclear to me currently where better to put this comment: Where would we put information about land fractions of grid cells? Still in a separate variable or somehow integrated into each of the data arrays we have? & what about the grid cells’ area? I’m interested in that because I often take grid cell area & land cover weighted spatial averages. I guess at least the grid cell area part is somehow covered by (@mathause’s) xarray functions?

ROADMAP.md
```diff
@@ -0,0 +1,76 @@
...
+- Define how the internal arrays in MESMER should be represented within xarray ([#105](https://github.com/MESMER-group/mesmer/issues/105)).
+- Rewrite the statistical calibration and emulation functions such that they work with xarray arrays.
+
+#### High level data structure
+
+Currently MESMER uses nested dictionaries as high level data structure. For example global mean temperature for a specific model and scenario is available as `GSAT["IPSL-CM6A-LR"]["ssp126"]`. This data structure is not very flexible (e.g. if there are no scenarios one level is superfluous while emulating a GCM-RCM chain might need another level, see [#113](https://github.com/MESMER-group/mesmer/issues/113)). A flat high level data structure would allow to circumvent these problems.
```
leabeusch, line 33: For local emulations it gets even worse. There, the variable name (i.e., `"tas"`) itself contributes yet another level of nestedness at the moment.

ROADMAP.md
```diff
@@ -0,0 +1,76 @@
...
+- Define an appropriate output format for each statistical task (see also the section _Persisting calibrated parameters_).
+
+Note that these tasks are highly dependent on the _Internal data structures_.
+
+
+### Persisting calibrated parameters
+
+Currently calibrated parameters are stored in a python-internal format (the [pickle](https://docs.python.org/3/library/pickle.html) format via [`joblib.dump`](https://joblib.readthedocs.io/en/latest/generated/joblib.dump.html)). The current format is problematic from a security perspective (see the [pickle documentation](https://docs.python.org/3/library/pickle.html)) and can pose problems with backward compatibility ([#50](https://github.com/MESMER-group/mesmer/issues/50)).
+
+The calibrated parameters (e.g. the spatially-explicit linear regression coefficients) need to be saved in a more standard format, i.e. as netCDFs ([#65](https://github.com/MESMER-group/mesmer/issues/65)).
```
leabeusch, line 56: Not just the calibrated paramters but also the emulations themselves! Altough we'd probably more often distribute calibrated paramters than emulations... still: it'd be a pity to keep the emulations in the clearly suboptimal pickle format after hopefully making it much more feasible to save them as netcdf too, through the planned internal data structure changes.

ROADMAP.md
```diff
@@ -0,0 +1,76 @@
...
+
+Specific tasks
+- Make sure the calibrated parameters for a given statistical step in MESMER are returned as xarray Dataset, which can directly be stored as netCDF.
+- Define how calibration-specific configuration should be saved (e.g. within the netCDF or external to the netCDF in a file.)
+- Define how to save global parameters that are calibration independent.
+
+### Configuration infrastructure
+
+The current configuration relies on a python file which includes all available settings. The settings are passed to each function, there is no default configuration, and the configuration file also calculates some values.
```
leabeusch, line 65: What do you mean by "there is no default configuration"? Just that we only have the one for the test there? But yes, I agree with everyone on this. Should clearly move to a better configuration format.

ROADMAP.md
```diff
@@ -0,0 +1,76 @@
...
+It would be beneficial to re work the current setup of the configuration. The configuration should be a static file (e.g. in yaml format) and default values should be set where appropriate.
+
+
+### Increase code test coverage
+
+Adding tests to the code ensures that the behavior of the code is as expected. It also guards against unintentional and unexpected changes when updating the code. Further, code tests can help to think about the code structure and design to make individual part as independent from each other as possible (else it's difficult to test). The current structure of MESMER makes it difficult to add tests because all the parts are very tightly linked.
+
+### Remove ETH specific code paths
+
+Code to load data is currently written to assuming the ETH data structure ([#61](https://github.com/MESMER-group/mesmer/issues/61)). At some point we should allow the users to hold the data however they want, load it how they want and then have interfaces into which they pass the data.
```
leabeusch, line 76: This probably goes a bit in the direction of "what is MESMER" again. I.e., what kind of data should be able to fed into MESMER eventually? Do first ideas on this need to be included in this roadmap too or is this just the roadmap of getting the MESMER we currently have into a better structure?

leabeusch (commented):
Here my first thoughts on this. Hopefully including some helpful ones. :)

ROADMAP.md
```diff
@@ -0,0 +1,76 @@
...
+Many of these changes are intricately connected with each other, therefore we think that rewriting large parts of MESMER is the most efficient way forward.
+
+## Envisioned updates
+
+### User interface
+
+Running MESMER currently involves a mixture of data loading, data preprocessing, defining the emulation model for calibration as well as for drawing realizations. We hope to improve on this by addressing the following points:
+
+- Clearer separation of individual steps from data loading and preprocessing.
+- More intuitive way to define a "emulation model".
```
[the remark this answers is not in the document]
mathause: Yes, I mean a statistical model. Maybe not the best choice of words. We should define some nomenclature.

ROADMAP.md
mathause: Should MESMER return arrays with `cell` as dim? And then it's a postprocessing step to get back to lat & lon? If we want to do it over all grid cells - do we still flatten them? (no need to answer)

yquilcaille: +1. We have ESMs for now, RCMs with @woodhome23, probably SCMs at some point, maybe IAMs one day, i've already had different statistical models... I think that we should avoid naming a dimension simply "model".

ROADMAP.md
yquilcaille: I would say having a separate variable for support information (cell area, land fraction, latitude, longitude, selection of land grid points, eventually stuff about regions, etc) could fit in a separate variable. Although, we could still have latitude and longitude in all items the DataList of inputs and one specific item in the list that would have only this support data?

ROADMAP.md
yquilcaille: Excellent question... And actually, what kind of data for calibration or for training?

Actually, a thought on that. Having MESMER on this github _with_ parameters already loaded here ensures that every user can try MESMER, run it with scenarios, etc, independently from the data for calibration. Not all users would want and/or have the data for another calibration. But it is nice to keep it possible.

Then I would suggest not only removing ETH specific code paths for training, so that if an user wants to go deep in it, it can. But also having a prepared set of parameters and a very accessible emulations part.

ROADMAP.md
mathause: No this comment is very relevant. I added https://github.com/MESMER-group/mesmer/issues/65#issuecomment-959123361 on this.

ROADMAP.md
mathause: We could have pre-defined settings for some stuff we want to change seldom, e.g. instead of setting 

https://github.com/MESMER-group/mesmer/blob/f86a87aad09162301b60c1b13ac0e90e1c6d95b0/configs/config_across_scen_T_cmip6ng_test.py#L53

"manually", it could be 

```python
mesmer.config["reference_period"]
>>> slice("1850", "1900")
```
per default.

ROADMAP.md
mathause: Very good point! We should definitely have this in the roadmap (maybe even more important than what we have there now).

ROADMAP.md
```diff
@@ -0,0 +1,78 @@
...
+Many of these changes are intricately connected with each other, therefore we think that rewriting large parts of MESMER is the most efficient way forward.
+
+## Envisioned updates
+
+### User interface
+
+Running MESMER currently involves a mixture of data loading, data preprocessing, defining the statistical model for calibration, storing the determined model parameters, as well as drawing realizations. We hope to improve on this by addressing the following points:
+
+
+- Clearer separation of individual steps from data loading and preprocessing.
```
mathause, line 15: TODO: add more points

mathause (commented):
It might be good to merge this before we submit the open research data proposal. So we can point to it from there. I read it again and tried to make some minor points either more or less specific - depending on your comments.

@woodhome23 It does currently not entirely align (in terms of structure) with what we write in the proposal. IMHO this should be ok - do you agree?

ROADMAP.md
woodhome23: yes thanks, that sounds good

ROADMAP.md
```diff
@@ -0,0 +1,79 @@
...
+- Define an appropriate output format for each statistical task (see also the section _Persisting calibrated parameters_).
+
+Note that these tasks are highly dependent on the _Internal data structures_.
+
+### Persisting calibrated parameters and emulations
+
+Currently, calibrated parameters and emulations are stored in a python-internal format (the [pickle](https://docs.python.org/3/library/pickle.html) format via [`joblib.dump`](https://joblib.readthedocs.io/en/latest/generated/joblib.dump.html)). The current format is problematic from a security perspective (see the [pickle documentation](https://docs.python.org/3/library/pickle.html)) and can pose problems with backward compatibility ([#50](https://github.com/MESMER-group/mesmer/issues/50)).
+
+The calibrated parameters (e.g. the spatially-explicit linear regression coefficients) and  need to be saved in a more standard format, i.e. as netCDFs ([#65](https://github.com/MESMER-group/mesmer/issues/65)). Note that switching the internal _Array data structure_ to xarray will greatly simplify this task because they can be directly saved as netCDF files.
```
mathause, line 59:
```suggestion
The calibrated parameters (e.g. the spatially-explicit linear regression coefficients) and need to be saved in a more standard format, i.e. as netCDFs ([#65](https://github.com/MESMER-group/mesmer/issues/65)). Note that switching the internal _Array data structure_ to xarray will greatly simplify this task because they can be directly saved as netCDF files.
```

PR #116, opened 2021-11-30
Linear regression

znicholls:
- [x] Tests added


CHANGELOG.rst
```diff
@@ -7,6 +7,9 @@ v0.9.0 - unreleased
 New Features
 ^^^^^^^^^^^^
 
+- Add ``mesmer.core.linear_regression``. Starts the process of refactoring the
```
mathause, line 10: Should it live in `mesmer.stat` or `mesmer.core.stat`?

tests/integration/test_core_integration.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def test_basic_regression_multidimensional():
+    res = mesmer.core.linear_regression.linear_regression(
+        [[0, 1], [1, 3], [2, 4]], [2, 7, 8]
+    )
+
+    # intercept before coefficients, in same order as columns of
+    # predictors
+    npt.assert_allclose(res, [-2, -3, 4], atol=1e-10)
```
mathause, line 28: Isn't `rtol` enough here? (I don't mind this as at all but I usually only set this if required).

tests/unit/test_core.py
```diff
@@ -0,0 +1,75 @@
...
+        mocked_linear_regression.return_value = mock_regressor
+
+        if weight is None:
+            expected_weights = None
+            res = mesmer.core.linear_regression.linear_regression(predictors, target)
+        else:
+            expected_weights = weight
+            res = mesmer.core.linear_regression.linear_regression(
+                predictors, target, weight
+            )
```
mathause, line 40: Is this if-else actually needed?

tests/unit/test_core.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+@pytest.mark.parametrize(
+    "predictors,target,weight",
+    (
+        ([[1], [2], [3]], [1, 2, 2], None),
+        ([[1, 2, 3], [2, 4, 0]], [1, 2], [3, 1]),
+    ),
+)
+def test_linear_regression(predictors, target, weight):
```
mathause, line 17: Ok I see now what you are doing here & that seems really interesting conceptually but it feels a bit over the top for me - the `test_basic_regression` should be enough? But I am maybe also missing your point?

tests/unit/test_core.py
```diff
@@ -0,0 +1,75 @@
...
+    # written without proper dependency injection. See e.g.
+    # https://stackoverflow.com/a/46865495 which recommends against this
+    # approach. At the moment, I can't see how to write a suitably simple
+    # function for regressions that uses proper dependency injection and
+    # doesn't make the interface more complicated.
+    mock_regressor = mock.Mock()
+    mock_regressor.intercept_ = 12
+    mock_regressor.coef_ = [123, -38]
+
+    with mock.patch(
```
mathause, line 28: Oh cool you managed to get `mock` to work. I gave up and used pytest's `monkeypatch` fixture.

mathause (commented):
Thanks for getting this started! Looks good but I don't entirely get why you split your tests in integration and unit (?) tests. In my opinion your integration tests are also unit tests. And they could IMHO all live in `test_linear_regression.py`. But maybe I am missing your point. Should we quickly discuss this over zoom?

mesmer/core/linear_regression.py
```diff
@@ -0,0 +1,29 @@
...
+def linear_regression(predictors, target, weights=None):
+    """
+    Perform a linear regression
+
+    Parameters
+    ----------
+    predictors : array-like of shape (n_samples, n_predictors)
+        Array of predictors
+
+    target : array-like of shape (n_samples,)
```
mathause, line 14: Should we allow `(n_samples,) or (n_samples, n_targets)`? Equal to what `LinearRegression().fit` does? If so you need to update the stacking.


```python
    intercept = np.atleast_2d(reg.intercept_).T
    coef = np.atleast_2d(b)

    return np.hstack([intercept, coef])
```

This would return a `(n_targets, n_predictors)` array. We could directly pass an a flattened 2D array, e.g. with dims `(time, cells)` to fit all grid points at once.


Some tests:

<details>

```python
# one predictor
x = np.array([[0], [1], [2]]) # (3, 1)
y = np.array([2, 7, 10]) # (3,) 

a = linear_regression(x, y)

assert a.shape == (1, 2)


x = np.array([[0], [1], [2]]) # (3, 1)
y = np.array([[2], [7], [10]]) # (3, 1)

a = linear_regression(x, y)

assert a.shape == (1, 2)


x = np.array([[0], [1], [2]]) # (3, 1)
y = np.array([[2, 4], [7, 14], [10, 20]]) # (3, 2)

a = linear_regression(x, y)

assert a.shape == (2, 2)


# two predictors
x = np.array([[0, 1], [1, 3], [2, 4]]) # (3, 2)
y = np.array([2, 7, 10]) # (3, )

a = linear_regression(x, y)

assert a.shape == (1, 3)


x = np.array([[0, 1], [1, 3], [2, 4]])  # (3, 2)
y = np.array([[2], [7], [10]])  # (3, 1)

a = linear_regression(x, y)

assert a.shape == (1, 3)


x = np.array([[0, 1], [1, 3], [2, 4]])  # (3, 2)
y = np.array([[2, 4], [7, 14], [10, 20]]) # (3, 2)

a = linear_regression(x, y)

assert a.shape == (2, 3)
```

</details>

CHANGELOG.rst
znicholls: Happy either way (toss a coin)

tests/integration/test_core_integration.py
znicholls: Good pick up, fixed

tests/unit/test_core.py
znicholls: Is just a way of checking what the defaults are (if `weights is None` then we don't pass `weight` in as an argument i.e. the default gets used, otherwise we pass `weight` and make sure it is passed correctly, it is just coincidence that in this case the default is also `None`, normally you'd have `expected_weights=1` or something else where it is more obvious that there is a default value). Option b would be to have two tests (`test_linear_regression_defaults` and `test_linear_regression_weight`) but I prefer to have the defaults and passing options tested next to each other just so I can see more easily what the difference is.

tests/unit/test_core.py
znicholls: It's definitely over the top right now. In a previous project (https://gitlab.com/netcdf-scm/netcdf-scm) I went hard on unit tests with lots of mocking and they were really helpful in the long-run because I could catch lots of issues without having to use real data all the time (or having my tests explode when iris made an update). I would encourage getting people into the habit of using mocking but if it's too over the top I'm happy to let it go.

mesmer/core/linear_regression.py
znicholls: I'm happy to do this. I wrote it with no stacking based on our previous call where you asked to have no stacking in the calls. However, forcing it to always be `(n_samples, n_targets)` isn't a bad idea as we can then fit many grid points at once as you say. I'll change now, thanks for the tests.

znicholls: > e.g. with dims (time, cells) to fit all grid points at once

Probably dims `(sample, cell)` given how bad our time handling is with this hist vs. scen business but yes that will be much more efficient.

znicholls:
> I don't entirely get why you split your tests in integration and unit (?) tests. In my opinion your integration tests are also unit tests

The distinction I make (and it's pretty arbitrary so happy to go with a different convention for MESMER) is that unit tests should never fail because of an external library i.e. they should only ever call things which have been mocked. Everything that is in integration actually calls the underlying sklearn package (so would fail if sklearn ever changed its API) while everything in unit mocks out sklearn so would pass even if sklearn completely re-wrote its API. 

> Should we quickly discuss this over zoom?

Happy to if you want. Also happy to just go with whatever your favourite convention is.

mathause (commented):
My first thought is to not distinguish between integration and unit tests and try to avoid mocks if possible. My second thought was - maybe I should give it a go so I am actually entitled to an opinion :smile: 

Can you point me to a test or two where you think it added a lot of benefit?

znicholls:
> Can you point me to a test or two where you think it added a lot of benefit?

An obvious one is here where you can test processing of responses without needing the internet https://gitlab.com/netcdf-scm/netcdf-scm/-/blob/master/tests/unit/test_citing.py#L176

This https://gitlab.com/netcdf-scm/netcdf-scm/-/blob/master/tests/unit/test_stitching.py#L232 allowed me to do a bunch of recursive path checking without actually needing to setup paths.

The other big advantage of mocking is that it forces you to write small functions. If it's not easy to mock, there's probably something wrong with your code.

mathause:
Thanks for the examples. I definitely agree with mocking web requests and can also see merit for mocking `glob`. Otherwise I am not so sure

> If it's not easy to mock, there's probably something wrong with your code.

I would rather say - if you have to use mock there's probably something wrong with your code :wink: E.g. instead of mocking glob I would try to write a pure function that takes a list of files as input, which you can then test to your heart's desire.

```python
def get_parent_file_path(path):
    files = glob.glob(path)
    return _get_parent_file_path(files)

def _get_parent_file_path(files):
    # do stuff
    return parent
```

znicholls:
> I would rather say - if you have to use mock there's probably something wrong with your code 😉 E.g. instead of mocking glob I would try to write a pure function that takes a list of files as input, which you can then test to your heart's desire.

That's a good example. In that case I would say that you test `_get_parent_file_path` however you want. Nonetheless, you still have the issue of how to test `get_parent_file_path`. In my experience there are two options: 

1. Test `get_parent_file_path` without mocking. In that case you are implicitly testing `_get_parent_file_path` at the same time. 
2. Test `get_parent_file_path` with mocking i.e. make sure that it calls glob and then passes the output to `_get_parent_file_path`. Test `_get_parent_file_path` separately to make sure it does the right thing with a list of file paths.

If there's only two layers the first one is almost always fine. However, if you have a command-line interface, which triggers off a whole bunch of other processes (say 50) then it quickly becomes impossible to test all the paths through all the different layers. In my experience, getting mocking into the testing early makes it much simpler to build the tests and be confident that all paths are covered. If I know the two paths through my first function work correctly, and the two paths through the second function work, and the two paths through my last function work, then I can conclude that all 8 combinations of paths will behave. However if I test the first function without mocking, I have to explicitly test that all 8 paths will behave correctly in order to be sure that this function is actually hooked up correctly. You can imagine that this quickly fails once it's 10 layers rather than 3. 

My two cents, let's discuss later.

znicholls:
@mathause the latest commits hopefully clean things up a bit. I've made #118 for when I get to writing docs.

mathause (approved):
Thanks for the PR and the discussion! Feel free to merge anytime.

PR #131, opened 2022-01-20
version fixes

mathause:
This PR does a number of things (maybe a bit too much but oh well).
- add `python_requires=">=3.7"` to setup.py - it's important to explicitly state this.
- update the CI:
  - merge `test` and `test-install` (and added a `Test installation` to the `test` CI) - I think we can merge these.
  - We now explicitly run the tests on different python versions
  - remove the py36 test and add a py39 test
  - move linting to py39
- Use mamba in the Makefile (if available)
- unpin xarray & recreate the test files
- some more f-strings

Still open: explicitly define & test minimal supported versions of packages. I would wait with this until we got rid of the pickle stuff (probably).


mathause:
Very strange - this picks up regionmask=0.5.0. No idea why.

znicholls (commented):
lgtm, unclear why it's failing

mathause:
It happens somewhere deep in the dependency tree... Some very indirect dependency of regionmask restricts `openssl<=1.1.1l` while python now gets openssp=3.0.0... I am pinning regionmask, this should work.

PR #134, opened 2022-02-23
add LinearRegression class

mathause:
- [ ] Closes #xxx
 - [x] Tests added

This adds a `LinearRegression` class. To fit a linear regression you would call:

```python
lr = mesmer.core.linear_regression.LinearRegression()
lr.fit({"tas": tas}, target, "time")
lr.params
```

This sets `lr.params` - i.e. the estimated intercept and slope for tas.

The next examples for one gridpoint ("cell") actually work - you can manually set `lt.params` and then call `lr.predict` and `lr.residuals` to get the predicted values and the residuals:

```python
import mesmer.core.linear_regression
import xarray as xr

# set some params (i.e. intercept and slope)
lr = mesmer.core.linear_regression.LinearRegression()

params = xr.Dataset(data_vars={"intercept": ("cell", [5]), "tas": ("cell", [0.1])})
lr.params = params

tas_predictor = xr.DataArray([0, 1, 2], dims="time")
target = xr.DataArray([[5, 8, 0]], dims=("cell", "time"))


predicted = lr.predict({"tas": tas_predictor})
residuals = lr.residuals({"tas": tas_predictor}, target)
```


mathause:
This is basically an xarray wrapper for `from sklearn.linear_model.LinearRegression`.

mathause:
@yquilcaille If you are interested you can have a look here. You can concentrate on the `linear_regression.py` file - the others are not super important.

tests/integration/test_linear_regression.py
```diff
@@ -9,10 +9,93 @@
 
 from .utils import trend_data_1D, trend_data_2D
 
-# TEST XARRAY WRAPPER
 
+def LinearRegression_fit_wrapper(*args, **kwargs):
+    # wrapper for LinearRegression().fit() because it has no return value - should it?
```
znicholls, line 14: I um and ah about this. I think it's probably a good idea to just get people used to the pattern of calling `fit` and then getting the `params` or whatever in the next line. That's what's used in sklearn and following that is probably a good idea. We could add a function like this to the codebase of course, it's so simple that it probably doesn't hurt...

znicholls (approved):
Looks sick, such a nice and elegant solution to how we store parameters, make the fitting and prediction stay close to each other etc. Not sure if there's changelog etc. to be done but I think this is a great pattern for us to follow

tests/integration/test_linear_regression.py
mathause: Yes, I recently heard somewhere that a class method should either return something _or_ change state but not both. So I will keep it as is.

mathause:
Thanks for the feedback! Will add a changelog and merge. I am not yet sure if the AR process will work you just as nice...

CHANGELOG.rst
```diff
@@ -6,7 +6,10 @@ v0.9.0 - unreleased
 
 New Features
 ^^^^^^^^^^^^
-
+- Create :py:class:`mesmer.core.linear_regression.LinearRegression` which encapsulates
+  ``fit``, ``predict``, etc. methods aroung linear regression
```
mathause, line 10:
```suggestion
  ``fit``, ``predict``, etc. methods around linear regression
```

mesmer/core/linear_regression.py
```diff
@@ -2,9 +2,168 @@
...
 import numpy as np
 import xarray as xr
-from sklearn.linear_model import LinearRegression
 
-from .utils import _check_dataarray_form
+from .utils import _check_dataarray_form, _check_dataset_form
+
+
+class LinearRegression:
+    """Ordinary least squares Linear Regression for xarray.DataArray objects."""
```
mathause, line 10:
```suggestion
    """Ordinary least squares Linear Regression for xr.DataArray objects."""
```

PR #139, opened 2022-03-31
add _fit_auto_regression functions

mathause:
- [x] Tests added

Adds `_fit_auto_regression_xr` and `_fit_auto_regression_np` - thin wrappers around `statsmodels.tsa.ar_model.AutoReg`.

In contrast to the linear regression I have:
- reworked the internal code by creating a dummy `DataArray` - I think that simplifies the code and looks pretty neat
- not written any 'numerical' tests - it only checks the shape etc. of the results

I am not sure we can follow exactly the same pattern as for the `LinearRegression` class because the coeffs are averaged over the scens. It would be nice to have the same pattern for both but I am not sure which will have to give...

@yquilcaille @znicholls

### DIFF
--- a/mesmer/calibrate_mesmer/train_gv.py
+++ b/mesmer/calibrate_mesmer/train_gv.py
@@ -196,7 +196,7 @@ def train_gv_AR(params_gv, gv, max_lag, sel_crit):
     AR_order_sel = int(np.percentile(AR_order_scens_tmp, **pct_kwargs))
 
     # determine the AR params for the selected AR order
-    res = list()
+    params_scen = list()
     for scen_idx, scen in enumerate(gv.keys()):
         data = gv[scen]
 
@@ -206,16 +206,16 @@ def train_gv_AR(params_gv, gv, max_lag, sel_crit):
         params = _fit_auto_regression_xr(data, dim="time", lags=AR_order_sel)
         params = params.mean("run")
 
-        res.append(params)
+        params_scen.append(params)
 
-    res = xr.concat(res, dim="scen")
-    res = res.mean("scen")
+    params_scen = xr.concat(res, dim="scen")
+    params_scen = params_scen.mean("scen")
 
     # TODO: remove np.float64(...) (only here so the tests pass)
     params_gv["AR_order_sel"] = AR_order_sel
-    params_gv["AR_int"] = np.float64(res.intercept.values)
-    params_gv["AR_coefs"] = res.coeffs.values.squeeze()
-    params_gv["AR_std_innovs"] = np.float64(res.standard_deviation.values)
+    params_gv["AR_int"] = np.float64(params_scen.intercept.values)
+    params_gv["AR_coefs"] = params_scen.coeffs.values.squeeze()
+    params_gv["AR_std_innovs"] = np.float64(params_scen.standard_deviation.values)
 
     # check if fitted AR process is stationary
     # (highly unlikely this test will ever fail but better safe than sorry)
--- a/mesmer/calibrate_mesmer/train_gv.py
+++ b/mesmer/calibrate_mesmer/train_gv.py
@@ -213,7 +213,7 @@ def train_gv_AR(params_gv, gv, max_lag, sel_crit):
 
     # TODO: remove np.float64(...) (only here so the tests pass)
     params_gv["AR_order_sel"] = AR_order_sel
-    params_gv["AR_int"] = np.float64(res.trend.values)
+    params_gv["AR_int"] = np.float64(res.intercept.values)
     params_gv["AR_coefs"] = res.coeffs.values.squeeze()
     params_gv["AR_std_innovs"] = np.float64(res.standard_deviation.values)
 
--- a/mesmer/calibrate_mesmer/train_gv.py
+++ b/mesmer/calibrate_mesmer/train_gv.py
@@ -12,8 +12,11 @@ import os
 import joblib
 import numpy as np
 import statsmodels.api as sm
+import xarray as xr
 from packaging.version import Version
-from statsmodels.tsa.ar_model import AutoReg, ar_select_order
+from statsmodels.tsa.ar_model import ar_select_order
+
+from mesmer.core.auto_regression import _fit_auto_regression_xr
 
 
 def train_gv(gv, targ, esm, cfg, save_params=True, **kwargs):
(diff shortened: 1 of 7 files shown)


mesmer/core/auto_regression.py
```diff
@@ -0,0 +1,77 @@
...
+        data,
+        input_core_dims=[[dim]],
+        output_core_dims=((), ("lags",), ()),
+        vectorize=True,
+        output_dtypes=[float, float, float],
+        kwargs={"lags": lags},
+    )
+
+    # TODO: are the names appropriate?
+    data_vars = {"trend": trend, "coeffs": coeffs, "standard_deviation": std}
```
mathause, line 39: Rename `trend` to `intercept`.

mesmer/calibrate_mesmer/train_gv.py
```diff
@@ -193,29 +196,26 @@ def train_gv_AR(params_gv, gv, max_lag, sel_crit):
...
+        params = params.mean("run")
+
+        res.append(params)
+
+    res = xr.concat(res, dim="scen")
+    res = res.mean("scen")
+
+    # TODO: remove np.float64(...) (only here so the tests pass)
+    params_gv["AR_order_sel"] = AR_order_sel
+    params_gv["AR_int"] = np.float64(res.trend.values)
```
mathause, line 216:
```suggestion
    params_gv["AR_int"] = np.float64(res.intercept.values)
```

tests/integration/test_auto_regression.py
```diff
@@ -0,0 +1,92 @@
...
+@pytest.mark.parametrize("lags", [1, 2])
+def test_auto_regression_xr_1D(lags):
+
+    data = trend_data_1D()
+    res = mesmer.core.auto_regression._fit_auto_regression_xr(data, "time", lags=lags)
+
+    _check_dataset_form(
+        res,
+        "_fit_auto_regression_result",
+        required_vars=["trend", "coeffs", "standard_deviation"],
```
mathause, line 29:
```suggestion
        required_vars=["intercept", "coeffs", "standard_deviation"],
```

tests/integration/test_auto_regression.py
```diff
@@ -0,0 +1,92 @@
...
+    )
+
+    _check_dataarray_form(res.trend, "trend", ndim=0, shape=())
```
mathause, line 32:
```suggestion
    _check_dataarray_form(res.intercept, "intercept", ndim=0, shape=())
```

tests/integration/test_auto_regression.py
```diff
@@ -0,0 +1,92 @@
...
+
+    data = trend_data_2D()
+    res = mesmer.core.auto_regression._fit_auto_regression_xr(data, "time", lags=lags)
+
+    (n_cells,) = data.cells.shape
+
+    _check_dataset_form(
+        res,
+        "_fit_auto_regression_result",
+        required_vars=["trend", "coeffs", "standard_deviation"],
```
mathause, line 52:
```suggestion
        required_vars=["intercept", "coeffs", "standard_deviation"],
```

tests/integration/test_auto_regression.py
```diff
@@ -0,0 +1,92 @@
...
+    )
+
+    _check_dataarray_form(res.trend, "trend", ndim=1, shape=(n_cells,))
```
mathause, line 55:
```suggestion
    _check_dataarray_form(res.intercept, "intercept", ndim=1, shape=(n_cells,))
```

mesmer/core/auto_regression.py
```diff
@@ -0,0 +1,77 @@
...
+        A ``xr.DataArray`` to estimate the auto regression over.
+    dim : str
+        Dimension along which to fit the auto regression over.
+    lags : int
+        The number of lags to include in the model.
+
+    Returns
+    -------
+    :obj:`xr.Dataset`
+        Dataset containing the estimated parameters of the ``trend``, the AR ``coeffs``
```
mathause, line 21:
```suggestion
        Dataset containing the estimated parameters of the ``intercept``, the AR ``coeffs``
```

mesmer/core/auto_regression.py
```diff
@@ -0,0 +1,77 @@
...
+    ----------
+    data : np.array
+        A numpy array to estimate the auto regression over. Must be 1D.
+    lags : int
+        The number of lags to include in the model.
+
+    Returns
+    -------
+    trend : :obj:`np.array`
+        Trend part of the fited AR model.
```
mathause, line 59:
```suggestion
    intercept : :obj:`np.array`
        Intercept of the fited AR model.
```

mesmer/core/auto_regression.py
```diff
@@ -0,0 +1,77 @@
...
+        data,
+        input_core_dims=[[dim]],
+        output_core_dims=((), ("lags",), ()),
+        vectorize=True,
+        output_dtypes=[float, float, float],
+        kwargs={"lags": lags},
+    )
+
+    # TODO: are the names appropriate?
+    data_vars = {"intercept": intercept, "coeffs": coeffs, "standard_deviation": std}
```
mathause, line 39: TODO: attach the order

mesmer/calibrate_mesmer/train_gv.py
```diff
@@ -193,29 +196,26 @@ def train_gv_AR(params_gv, gv, max_lag, sel_crit):
     AR_order_sel = int(np.percentile(AR_order_scens_tmp, **pct_kwargs))
 
     # determine the AR params for the selected AR order
-    params_gv["AR_int"] = 0
-    params_gv["AR_coefs"] = np.zeros(AR_order_sel)
-    params_gv["AR_order_sel"] = AR_order_sel
-    params_gv["AR_std_innovs"] = 0
-
+    res = list()
```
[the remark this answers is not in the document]
znicholls: Yep although if the function is small enough it will be clear from the context

mesmer/core/auto_regression.py
znicholls: I'm always in favour of full names so coefficients rather than coeffs but I don't mind too much

mesmer/core/auto_regression.py
```diff
@@ -0,0 +1,77 @@
...
+    -------
+    intercept : :obj:`np.array`
+        Intercept of the fitted AR model.
+    coeffs : :obj:`np.array`
+        Coefficients if the AR model. Will have as many entries as ``lags``.
+    std :obj:`np.array`
+        Standard deviation of the residuals.
+    """
+
+    from statsmodels.tsa.ar_model import AutoReg
```
znicholls, line 66: Why here and not wrapped in try except in top level? Is that an xarray pattern?

znicholls (approved):
Beautiful

znicholls:
> I am not sure we can follow exactly the same pattern as for the LinearRegression class because the coeffs are averaged over the scens. It would be nice to have the same pattern for both but I am not sure which will have to give

Hmm this is indeed tricky. Can we split into two steps: calculate coeffs for each scenario separately, then average over as a second? That might help us to have the same internal patterns, even if the interfaces do different stuff

mesmer/calibrate_mesmer/train_gv.py
```diff
@@ -193,29 +196,26 @@ def train_gv_AR(params_gv, gv, max_lag, sel_crit):
...
     for scen_idx, scen in enumerate(gv.keys()):
-        nr_runs = gv[scen].shape[0]
-        AR_order_runs_tmp = np.zeros(nr_runs)
-        AR_int_tmp = 0
-        AR_coefs_tmp = np.zeros(AR_order_sel)
-        AR_std_innovs_tmp = 0
+        data = gv[scen]
```
znicholls, line 201: If we split out a function for this inner loop, we might be able to make the pattern look more like linear regression does

znicholls: Although I guess it's almost as small as it can be already, the inner function would have to be called something like `_fit_auto_regression_with_mean_over_runs` I guess.

(Side note, what does 'run' mean here? Is that ensemble member?)

mesmer/core/auto_regression.py
mathause: I did this for the linear regression because the sklearn class has the same name as ours (`LinearRegression`) so I just followed the same pattern.

mesmer/calibrate_mesmer/train_gv.py
mathause: I think this is definitely a good idea. However, I am still not sure what our "outer" data structure should be (I have a prototype of a `DataList` but I am not entirely convinced).

run is an ensemble member. I just followed Lea's terminology. But I agree we should standardize this stuff...

mathause:
Thanks for the feedback. I am still very unsure how this whole thing should look like in the end but I think these changes make sense regardless... So I would probably merge this more or less as is and try to refactor the next chunk and hope this helps me to see some patterns...

I have a slight preference towards shorter names (as long as their meaning is clear, which is subjective so maybe I should just use the long ones anyway :-P).

PR #143, opened 2022-04-26
refactor stack_predictors_and_targets

mathause:
Yet another preparation PR for #141 - rewriting & refactoring `train_l_prepare_X_y_wgteq`. I first wanted to to it quick and dirty but https://github.com/MESMER-group/mesmer/pull/141#discussion_r854703265 by @znicholls prompted me to have a closer look at the function. It's not going to be fun to merge this into #141 but I can worry about this later.

In this PR I

- split the original function into two:
  - `get_scenario_weights`
  - `stack_predictors_and_targets`
- write a helper function to stack predictors and targets
- the stacked arrays are now kept in a dict (instead of an array with an additional dimension)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,10 @@ New Features
 Breaking changes
 ^^^^^^^^^^^^^^^^
 
+- The :py:func:`train_l_prepare_X_y_wgteq` was refactored and split into two functions:
+  :py:func:`get_scenario_weights` and :py:func:`stack_predictors_and_targets`
+  (`#143 <https://github.com/MESMER-group/mesmer/pull/143>`_).
+  By `Mathias Hauser <https://github.com/mathause>`_.
 
 Deprecations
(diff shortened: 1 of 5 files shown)


CHANGELOG.rst
```diff
@@ -20,6 +20,10 @@ New Features
 Breaking changes
 ^^^^^^^^^^^^^^^^
 
+- The :py:func:`train_l_prepare_X_y_wgteq` was refactored and split into two functions:
```
mathause, line 23:
```suggestion
- Refactor and split :py:func:`train_l_prepare_X_y_wgteq` into two functions:
```

mathause:
And no tests because I think these will not be the final functions[^1], but this should happen in the high level data structure. Well maybe it would be worth it anyway and I am just being lazy.

[^1]: However, I think it's worth it anyway because the refactored functions are (should be) easier to understand.

PR #160, opened 2022-06-11
Replaced sum with np.sum for improved performance

### LINKED ISSUE #154: micro-optimization: use np.sum instead of sum
Using `sum(data)` is much slower than `np.sum(data)` or `data.sum()`.

- `create_emus_OLS_each_gp_sep`:

https://github.com/MESMER-group/mesmer/blob/3892e764dd80416137a0bf584f316aa48ca1e11b/mesmer/create_emulations/create_emus_lt.py#L202

Note: would be superseded by #155

- `create_emus_lv_OLS`:

https://github.com/MESMER-group/mesmer/blob/3892e764dd80416137a0bf584f316aa48ca1e11b/mesmer/create_emulations/create_emus_lv.py#L273

Note: would be superseded by #156

- There may be more

nikhilnarla:
- [x] Closes #154

### DIFF
--- a/mesmer/create_emulations/create_emus_lt.py
+++ b/mesmer/create_emulations/create_emus_lt.py
@@ -199,6 +199,6 @@ def create_emus_OLS_each_gp_sep(params_lt, preds_lt, scen):
                 for pred in params_lt["preds"]
             ]
 
-            emus_lt[targ][:, gp] = sum(pred_vals) + params_lt["intercept"][targ][gp]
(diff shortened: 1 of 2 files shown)


mathause:
Thanks for the PR. Unfortunately, the tests failed - do you need help to debug this?

mathause:
As a hint - can you compare 

```python
import numpy as np

a = np.random.randn(5, 3)

sum(a)
np.sum(a)
```

PR #161, opened 2022-06-14
add _draw_auto_regression_np

### LINKED ISSUE #150: auto regression: add numpy-based 'predict' function
Extract the code to 'predict' auto regression and write a numpy-based function. This is used in two places:

- https://github.com/MESMER-group/mesmer/blob/9def7422e2b0b77a105a6c1443e3d7c1eb6b0858/mesmer/create_emulations/create_emus_lv.py#L112
- https://github.com/MESMER-group/mesmer/blob/9def7422e2b0b77a105a6c1443e3d7c1eb6b0858/mesmer/create_emulations/create_emus_gv.py#L111

The two approaches are slightly different - one is based on a multivariate gaussian and the other on a univariate. However, it should be possible to unify them. My idea would be to do both using `multivariate_normal` and then `squeeze` the 1D case.

```python
import numpy as np

np.random.seed(1)
a = np.random.normal(loc=0, scale=1, size=3)

np.random.seed(1)
b = np.random.multivariate_normal([0], [[1]], size=3)

np.testing.assert_allclose(a, b.squeeze())
```

mathause:
- [x] Closes #150
 - [x] Tests added

---

A first stab at factoring out 'predicting' an AR process. It does not look very clean to me at the moment but it works.

Comments
- The coefs need to be 2D with shape `n_coefs x n_gridpoints`
- The `covariance` must be passed and not the `standard_deviation`
- `emus_lv_tmp` should be renamed
- `emus_lv_tmp` has shape `n_emus x (n_ts + buffer) x n_cells` - maybe it would be better if this was `n_emus x n_cells x (n_ts + buffer)` (so we can use `@` to multiply and sum, this would require to transpose the coefs)

### DIFF
--- a/mesmer/core/auto_regression.py
+++ b/mesmer/core/auto_regression.py
@@ -2,25 +2,30 @@ import numpy as np
 import xarray as xr
 
 
-def _draw_auto_regression_np(
+def _draw_auto_regression_correlated_np(
     *, intercept, coefs, covariance, n_samples, n_ts, seed, buffer
 ):
     """
     Draw time series of an auto regression process with possibly spatially-correlated
     innovations
 
+    Creates `n_samples` auto-correlated time series of order `ar_order` and length
+    `n_ts` for each set of `n_coefs` coefficients (typically one set for each grid
+    point), the resulting array has shape n_samples x n_ts x n_coefs. The innovations
+    can be spatially correlated.
+
     Parameters
     ----------
-    intercept : ndarray, float
-        Intercept of the model. Must be scalar or have size n_cells.
-    coefs : ndarray
-        The coefficients of the autoregressive process. Must be a 2D array with shape
-        ar_order x n_cells, i.e., the autoregressive coefficients are aligned along
-        axis=0, while axis=1 contains all grid cells.
-    covariance : float or ndarray
-        The (co-)variance array. Needs to have shape n_cells x n_cells.
+    intercept : float or ndarray of length n_coefs
+        Intercept of the model.
+    coefs : ndarray of shape ar_order x n_coefs
+        The coefficients of the autoregressive process. Must be a 2D array with the
+        autoregressive coefficients along axis=0, while axis=1 contains all idependent
+        coefficients.
+    covariance : float or ndarray of shape n_coefs x n_coefs
+        The (co-)variance array. Must be symmetric and positive-semidefinite.
     n_samples : int
-        Number of samples to draw.
+        Number of samples to draw for each set of coefficients.
     n_ts : int
         Number of time steps to draw.
     seed : int
@@ -33,10 +38,12 @@ def _draw_auto_regression_np(
     -------
     out : ndarray
         Drawn realizations of the specified autoregressive process. The array has shape
-        n_samples x n_ts x n_cells.
+        n_samples x n_ts x n_coefs.
 
     Notes
     -----
+    The 'innovations' is the error or noise term.
+
     As this is not a deterministic function it is not called `predict`. "Predicting"
     an autoregressive process does not include the innovations and therefore asymptotes
     towards a certain value (in contrast to this function).
@@ -44,23 +51,22 @@ def _draw_auto_regression_np(
     intercept = np.array(intercept)
     covariance = np.atleast_2d(covariance)
 
-    # coeffs assumed to be ar_order x n_cells
-    ar_order, n_cells = coefs.shape
+    # coeffs assumed to be ar_order x n_coefs
+    ar_order, n_coefs = coefs.shape
 
-    # TODO: allow arbitrary lags? (e.g., [1, 12]) -> need to pass `ar_lags`
-    # see https://github.com/MESMER-group/mesmer/issues/164
+    # TODO: allow arbitrary lags? (e.g., [1, 12]) -> need to pass `ar_lags` (see #164)
     ar_lags = np.arange(1, ar_order + 1, dtype=int)
 
-    # ensure reproducibility (TODO: clarify approach to this see #35)
+    # ensure reproducibility (TODO: clarify approach to this, see #35)
     np.random.seed(seed)
 
     innovations = np.random.multivariate_normal(
-        mean=np.zeros(n_cells),
+        mean=np.zeros(n_coefs),
         cov=covariance,
         size=[n_samples, n_ts + buffer],
     )
 
-    out = np.zeros([n_samples, n_ts + buffer, n_cells])
+    out = np.zeros([n_samples, n_ts + buffer, n_coefs])
     for t in range(ar_order + 1, n_ts + buffer):
 
         ar = np.sum(coefs * out[:, t - ar_lags, :], axis=1)
--- a/mesmer/core/auto_regression.py
+++ b/mesmer/core/auto_regression.py
@@ -2,6 +2,37 @@ import numpy as np
 import xarray as xr
 
 
+def _predict_auto_regression_np(
+    intercept, coefs, covariance, n_emus, n_ts, n_cells, seed, buffer
+):
+
+    intercept = np.array(intercept)
+    covariance = np.atleast_2d(covariance)
+
+    # ensure reproducibility
+    np.random.seed(seed)
+
+    innovs = np.random.multivariate_normal(
+        mean=np.zeros(n_cells),
+        cov=covariance,
+        size=[n_emus, n_ts + buffer],
+    )
+
+    emus_lv_tmp = np.zeros([n_emus, n_ts + buffer, n_cells])
+
+    coefs = coefs
+    AR_order = coefs.shape[0]
+    ar_lags = np.arange(1, AR_order + 1, dtype=int)
+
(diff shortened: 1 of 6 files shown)


mesmer/create_emulations/create_emus_lv.py
```diff
@@ -180,39 +182,29 @@ def create_emus_lv_AR1_sci(emus_lv, params_lv, preds_lv, cfg):
...
-                    + params_lv["AR1_coef"][targ] * emus_lv_tmp[:, t - 1, :]
-                    + innovs[:, t, :]
-                )
-            emus_lv_tmp = emus_lv_tmp[:, buffer:, :]
-
-            print("Create the full local variability emulations")
-            emus_lv[scen][targ] += emus_lv_tmp
 
+            emus_lv[scen][targ] += emus_ar.squeeze()
+            print(f"{emus_lv[scen][targ]=}")
```
mathause, line 207:
```suggestion
```

mathause:
@yquilcaille - the PR is not finished yet but if you want to take a look at it, that'd be welcome.

mesmer/create_emulations/create_emus_lv.py
```diff
@@ -180,39 +182,28 @@ def create_emus_lv_AR1_sci(emus_lv, params_lv, preds_lv, cfg):
...
-                size=[nr_emus_v, nr_ts_emus_stoch_v + buffer],
-            )
-
-            print(
-                "Compute the contribution to emus_lv by the AR(1) process with the "
-                "spatially correlated innovations"
+            emus_ar = _predict_auto_regression_np(
+                intercept=intercept,
+                coefs=coefs,
+                covariance=covariance,
```
mathause, line 198:
```suggestion
                intercept=params_lv["AR1_int"][targ],
                # reshape to n_coefs x n_cells
                coefs=params_lv["AR1_coef"][targ][np.newaxis, :],
                covariance=params_lv["loc_ecov_AR1_innovs"][targ],
```

mesmer/create_emulations/create_emus_lv.py
```diff
@@ -180,39 +182,28 @@ def create_emus_lv_AR1_sci(emus_lv, params_lv, preds_lv, cfg):
...
             if len(emus_lv[scen]) == 0:
-
                 emus_lv[scen][targ] = np.zeros(nr_emus_v, nr_ts_emus_stoch_v, nr_gps)
 
-            # ensure reproducibility
-            np.random.seed(seed)
+            intercept = params_lv["AR1_int"][targ]
+            # reshape to n_coefs x n_cells
+            coefs = params_lv["AR1_coef"][targ][np.newaxis, :]
+            covariance = params_lv["loc_ecov_AR1_innovs"][targ]
```
mathause, line 190:
```suggestion
```

mesmer/create_emulations/create_emus_gv.py
```diff
@@ -157,33 +159,24 @@ def create_emus_gv_AR(params_gv, nr_emus_v, nr_ts_emus_v, seed):
...
-            + innovs_emus_gv[:, t]
-        )
-
-    # remove buffer
-    emus_gv = emus_gv[:, buffer:]
-
-    return emus_gv
+    emus_gv = _predict_auto_regression_np(
+        intercept=ar_int,
+        coefs=ar_coefs[:, np.newaxis],
```
mathause, line 173:
```suggestion
        # reshape to n_coefs x n_cells
        coefs=ar_coefs[:, np.newaxis],
```

mathause:
Marking this ready for review. @znicholls do you want to have a look?

mesmer/core/auto_regression.py
```diff
@@ -2,6 +2,72 @@
...
+        Drawn realizations of the specified autoregressive process. The array has shape
+        n_samples x n_ts x n_cells.
+
+    Notes
+    -----
+    As this is not an deterministic function it is not called `predict`. "Predicting"
+    an autoregressive process does not include the innovations and therefore asymptotes
+    towards certain value (in constrast to this function).
+    """
+
```
znicholls, line 42:
```suggestion
```

mesmer/core/auto_regression.py
```diff
@@ -2,6 +2,72 @@
...
+    intercept = np.array(intercept)
+    covariance = np.atleast_2d(covariance)
+
+    # coeffs must be ar_order x n_cells
```
znicholls, line 46:
```suggestion
    # coeffs assumed to be ar_order x n_cells
```

mesmer/core/auto_regression.py
```diff
@@ -2,6 +2,72 @@
...
+    ar_order, n_cells = coefs.shape
+
+    # TODO: allow arbitrary lags? (e.g., [1, 12]) -> need to pass `ar_lags`
```
znicholls, line 49: Make an issue and reference in comment?

mesmer/core/auto_regression.py
```diff
@@ -2,6 +2,72 @@
...
+    ar_lags = np.arange(1, ar_order + 1, dtype=int)
+
+    # ensure reproducibility
```
znicholls, line 52:
```suggestion
    # ensure reproducibility (TODO: clarify approach to this see #35)
```

mesmer/core/auto_regression.py
```diff
@@ -2,6 +2,72 @@
 import xarray as xr
 
 
+def _draw_auto_regression_np(
```
znicholls, line 5: Can we change the name given that this is not really an auto-regression draw (it's auto-regression plus the spatial covariance)? Perhaps `_draw_auto_regression_with_innovations_np`?

mesmer/core/auto_regression.py
```diff
@@ -2,6 +2,72 @@
...
+    *, intercept, coefs, covariance, n_samples, n_ts, seed, buffer
+):
+    """draw time series of an autoregressive process
```
znicholls, line 8:
```suggestion
    """Draw time series of an autoregressive process plus innovations
```

Can we add a one line definition of 'innovations' in here (or a reference to somewhere which explains this term, I've never seen it before)

mesmer/core/auto_regression.py
```diff
@@ -2,6 +2,72 @@
...
+    np.random.seed(seed)
+
+    innovations = np.random.multivariate_normal(
+        mean=np.zeros(n_cells),
+        cov=covariance,
+        size=[n_samples, n_ts + buffer],
+    )
+
+    out = np.zeros([n_samples, n_ts + buffer, n_cells])
+    for t in range(ar_order + 1, n_ts + buffer):
```
znicholls, line 62: Is there no way to vectorise this (maybe not, just seems odd for us to have loops)? Can also be done in a follow up PR

znicholls (approved):
What a beauty! A few comments but happy for you to make the call about whether you just merge this and create issues for anything that is tricky to address or just address things now.

mathause (commented):
Thanks for the fast review!

---

The 'innovations' in a auto regression is the error or noise term. I think the idea behind the name is that for each time step this is the 'new' element i.e. the _innovation_ - i.e. in 

`$x_t = x_{t - 1} * \phi + \epsilon_t$`

it would be the `\epsilon_t$`.

I'll add something to the docstring.

---

I have unified the the global (= 1 cell without correlation, obviously) and local (= n cells with auto correlation) case. So it's only "possibly" correlated. I am not sure if it is better to have one or two functions.

```python
class AutoRegression:

    def draw(...): ...

# OR

    def draw(...): ...

    def draw_correlated(...): ...

# OR

    def draw_1d(...): ...

    def draw_2d(...): ...

# OR ???

```

znicholls:
> I am not sure if it is better to have one or two functions

I would go one and just add something like the following to the description of the `coeff` parameter: "If `coeff` is set to zero then this is just a standard sample from an autoregressive process." That will help users know that this function can be used without innovations if they want. If we get more complicated, we might want to refactor but for now it should be fine.

mesmer/core/auto_regression.py
```diff
@@ -2,6 +2,72 @@
 import xarray as xr
 
 
+def _draw_auto_regression_np(
+    *, intercept, coefs, covariance, n_samples, n_ts, seed, buffer
+):
+    """draw time series of an autoregressive process
+
+    Parameters
+    ----------
```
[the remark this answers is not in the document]
znicholls: Add a docstring of what expected args are (there's a `*` in the function signature, but the docstring doesn't describe what is expected or where these args go). So the docstring would look like something like

```
Parameters
---------------
args
    Arguments passed through to something
```

Looking at it more now, the `*` in the function signature isn't used so could be redundant?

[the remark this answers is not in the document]
mathause: This indicates that all arguments are keyword only.

mathause (commented):
I renamed the function, added an explanatory sentence and tried to clear up the names. `n_cells` is now called `n_coefs` - I hope this is not confused with the order of the AR process.

znicholls:
Very nice. Other option would be `n_gridpoints`. I'm happy with whatever so once you choose what you like best, merge away

PR #169, opened 2022-06-23
setup options to setup.cfg

mathause:

I wasn't sure what to tackle next so I did some infrastructure maintenance :upside_down_face:

@znicholls what does the `TestCommand` do? Is that something that you use? Is used? Could that be removed or something similar be achieved via `conftest.py`?

The next steps would be to use `setuptools_scm` instead of `versioneer` and add a pyproject.toml file.


mathause:
Thanks for the feedback. Feels like I achieved something :upside_down_face:

PR #183, opened 2022-07-15
move gaspari_cohn and calc_geodist_exact to core

### LINKED ISSUE #158: move gaspari_cohn and calc_geodist_exact
`gaspari_cohn` and `calc_geodist_exact` are in the file `load_constant_files.py`:

https://github.com/MESMER-group/mesmer/blob/3892e764dd80416137a0bf584f316aa48ca1e11b/mesmer/io/load_constant_files.py#L23

https://github.com/MESMER-group/mesmer/blob/3892e764dd80416137a0bf584f316aa48ca1e11b/mesmer/io/load_constant_files.py#L75

This is the wrong place for the two functions and they should be moved to a better home. I don't have a super strong opinion where they need to go - probably `mesmer/core/computation.py`.

yquilcaille:
- [x] Closes #158 
 - [ ] Tests added

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -34,6 +34,9 @@ Breaking changes
   :py:func:`get_scenario_weights` and :py:func:`stack_predictors_and_targets`
   (`#143 <https://github.com/MESMER-group/mesmer/pull/143>`_).
   By `Mathias Hauser <https://github.com/mathause>`_.
+- Moved ``gaspari_cohn`` & ``calc_geodist_exact`` from ``io.load_constant_files`` to ``core.computation``
+  (`#158 <https://github.com/MESMER-group/mesmer/issues/158>`_).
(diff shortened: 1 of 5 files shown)


mathause:
Ok let's merge this and worry about its beauty later.

mathause (approved):

mathause:
Thanks @yquilcaille!

znicholls:
> * is `computation` the right place?

Good enough for now

> * it is now a hard deprecation? Should we warn users that the files moved? (I tend towards no)

In future, once we have something semi-stable, we should get in the habit of deprecation. Given how unstable things are right now, I agree that it's not worth it.

> * we may want to rename `gaspari_cohn` to `_gaspari_cohn_np` etc. (but maybe in another PR)

Yep follow up PR

> @znicholls do you want to to take a look?

Good call just merging when I am slow

PR #184, opened 2022-08-03
refactor localized covariance

### LINKED ISSUE #187: do not `allow_singular=True` in `multivariate_normal.logpdf`?
In the function to find the optimal localization radius we calculate the logpdf of a multivariate normal distribution and allow singular matrices:

https://github.com/MESMER-group/mesmer/blob/f5f7be1720c2794f2ad23aa39c139f2291226399/mesmer/calibrate_mesmer/train_lv.py#L386-L388

(note that this code is currently being moved around in #184). However, this is problematic because it can distort the (negative) log-likelihood.

<details>

```python

import numpy as np
from scipy.stats import multivariate_normal

np.random.seed(0)
data = np.random.rand(5, 3)
data_train = data[1::2]
data_cv = data[::2]

out = list()
for crosscov in np.arange(0, 1.01, 0.01):

    localizer = np.full((3, 3), fill_value=crosscov)
    localizer[np.diag_indices(3)] = 1


    cov = np.cov(data_train, rowvar=False)

    log_likelihood = multivariate_normal.logpdf(
        data_cv, cov=cov * localizer, allow_singular=True
    ).sum()
    out.append(-log_likelihood)

out = np.array(out)
```

</details>

singular

In the example the last one leads to a singular matrix - which leads to the smallest negative log likelihood, which would then be selected.

This problem is mitigated because, mesmer aborts the search early, i.e. as soon as the negative-log-likelihood starts to increase (hard to see but this would be the fourth element in out [`np.argmin(out[:-1])` returns 3]).

Still I am a bit worried that we may end up selecting the wrong localization radius because of this & reluctant to leave something in that is obviously wrong.

---

Options

1. Remove the localization radius if a singular matrix is detected for any crossvalidation fold.
2. Only skip the localization radius if all crossvalidation folds produce singular matrices. Instead of summing all negative-log-likelihoods we would need to average them. However, I don't know if this is something that is expected to often happen. (this may also have the advantage to make them better comparable when using a different number of folds. The disadvantage is that the negative log-likelihood is usually summed and not averaged).

cc @jschwaab @leabeusch

mathause:
- [x] Part of #153
 - [x] Closes  #187
 - [x] Tests added

This is a really tough one. My refactoring is all over the place :unamused: but I need to get it to github now to see it from another perspective :upside_down_face:

The tests are going to be nightmare - not looking forward to that.


mesmer/core/localized_covariance.py
```diff
@@ -0,0 +1,217 @@
...
+from scipy.stats import multivariate_normal
+
+
+def _adjust_ecov_ar1_np(covariance, ar_coefs):
+    """
+    adjust localized empirical covariance matrix for autoregressive process of order 1
+
+    Parameters
+    ----------
+    covariance : 2D np.array
```
mathause, line 11: I am not sure if I should name this `covariance` or `cov`. I thought `covariance` is more explicit but then I remembered that numpy uses `cov` (e.g. for `np.random.multivariate_normal`).

mesmer/core/localized_covariance.py
```diff
@@ -0,0 +1,217 @@
...
+def _find_localized_empirical_covariance_np(data, weights, localizer, k_folds):
+    """determine localized empirical covariance by cross validation
+
+    Parameters
+    ----------
+    data : 2D array
+        Data array with shape nr_samples x nr_gridpoints.
+    weights : 1D array
+        Weights for the individual samples.
+    localizer : dict of array-like
```
mathause, line 65: I tried to make the name a bit more general than `phi_gc` but I am not really happy with the name - ideas?

mesmer/core/localized_covariance.py
```diff
@@ -0,0 +1,217 @@
...
+        Dictonary containing the localization radii as keys and the localisation as
+        values. The localization must be 2D and of shape nr_gridpoints x nr_gridpoints.
+    k_folds : int
+        Number of folds to use for cross validation.
+
+    Returns
+    -------
+    localisation_radius : float
+        Selected localisation radius.
+    ecov : ndarray
```
mathause, line 75: `empirical_covariance` is a bit too long for my liking but do we actually need `empirical` in the name?

mesmer/core/localized_covariance.py
```diff
@@ -0,0 +1,217 @@
...
+        ecov = np.cov(data_est, rowvar=False, aweights=weights_est)
+        loc_ecov = localizer[localisation_radius] * ecov
+
+        # sum log likelihood of all crossvalidation folds
+        log_likelihood += _get_neg_loglikelihood(data_cv, loc_ecov, weights_cv)
+
+    return log_likelihood
+
+
+def _get_neg_loglikelihood(data, covariance, weights):
```
mathause, line 140: Maybe this should not be a separate function.

mesmer/core/localized_covariance.py
```diff
@@ -0,0 +1,217 @@
...
+        k_folds=k_folds,
+    )
+
+    ecov = np.cov(data, rowvar=False, aweights=weights)
+    loc_ecov = localizer[localisation_radius] * ecov
+
+    return localisation_radius, ecov, loc_ecov
+
+
+def _ecov_crossvalidation(localisation_radius, *, data, weights, localizer, k_folds):
```
mathause, line 108: I don't like the name very much (but then again it's very private, so not important).

mathause (commented):
@znicholls this is still super far from complete but maybe you can take a look at my bikeshedding questions :upside_down_face:

mesmer/core/localized_covariance.py
```diff
@@ -0,0 +1,227 @@
...
+
+    .. [3] Cressie, N. and Wikle, C. K.: Statistics for spatio-temporal data, John Wiley
+       & Sons, Hoboken, New Jersey, USA, 2011.
+    """
+
+    reduction_factor = np.sqrt(1 - ar_coefs**2)
+    reduction_factor = np.atleast_2d(reduction_factor)  # so it can be transposed
+
+    # equivalent to ``diag(reduction_factor) @ ecov @ diag(reduction_factor)``
+    return reduction_factor * reduction_factor.T * covariance
```
znicholls, line 53:
```suggestion
    return diag(reduction_factor) @ covariance @ diag(reduction_factor)
```

Is there some reason we don't do this?

mesmer/core/localized_covariance.py
```diff
@@ -0,0 +1,227 @@
...
+def _adjust_ecov_ar1_np(covariance, ar_coefs):
+    """
+    adjust localized empirical covariance matrix for autoregressive process of order 1
+
+    Parameters
+    ----------
+    covariance : 2D np.array
+        Empirical covariance matrix.
+    ar_coefs : 1D np.array
+        The coefficients of the autoregressive process of order 1.
```
znicholls, line 14:
```suggestion
        The coefficients of the autoregressive process of order 1.
        Must have length equal to the size of `covariance` (`covariance` is square so it doesn't matter which dimension you use)
```

Could also add this assertion in the function if I have understood this correctly (I may not have)

mesmer/core/localized_covariance.py
znicholls: Hmm tricky, particularly as I don't really know what I'm looking at. I'd suggest sticking with whatever name you like, and maybe adding references/latex into the Notes section if you think there's extra info which could be very helpful

mesmer/core/localized_covariance.py
```diff
@@ -0,0 +1,227 @@
...
+    and a leave-one-out cross validation otherwise.
+    """
+
+    if not isinstance(k_folds, int) or k_folds <= 1:
+        raise ValueError("'k_folds' must be an integer larger than 1.")
+
+    localisation_radii = sorted(localizer.keys())
+
+    # find _local_ minimum because
+    # experience tells: once stop selecting larger localisation radii, will not
```
znicholls, line 92:
```suggestion
    # experience tells: once we stop selecting larger localisation radii, we will not
```

mesmer/core/localized_covariance.py
```diff
@@ -0,0 +1,227 @@
...
+    weights : 1D array
+        Sample weights
+
+    Returns
+    -------
+    weighted_log_likelihood : float
+
+    Notes
+    -----
+    The mean is zero for all points.
```
znicholls, line 161:
```suggestion
    The mean is assumed to be zero for all points.
```

mesmer/core/localized_covariance.py
```diff
@@ -0,0 +1,227 @@
...
+        res = func(element, **kwargs)
+
+        if np.isinf(np.abs(res)):
+            raise ValueError("`fun` returned `+/-inf`")
+        elif res < current_min:
+            current_min = res
+        else:
+            return sequence[i - 1]
+
+    # warn if the local minimum is not reached?
```
znicholls, line 226: I'd add this warning

tests/integration/test_localized_covariance.py
```diff
@@ -0,0 +1,76 @@
...
+
+from mesmer.core.localized_covariance import (
+    _adjust_ecov_ar1_np,
+    _minimize_local_discrete,
+)
+
+
+def test_ecov_crossvalidation():
+
+    import numpy as np
```
znicholls, line 12:
```suggestion
```

Done above

tests/integration/test_localized_covariance.py
```diff
@@ -0,0 +1,76 @@
...
+
+    from mesmer.core.localized_covariance import _ecov_crossvalidation
```
znicholls, line 14: Move to top?

tests/integration/test_localized_covariance.py
```diff
@@ -0,0 +1,76 @@
...
+
+    # nontrivial localizer (symmetric and diag == 1) between 0..1
+    loc = np.random.uniform(size=(3, 3))
+    ls = loc * loc.T
+    ls[np.diag_indices(3)] = 1
+    localizer = {250: ls}
+
+    _ecov_crossvalidation(
+        250, data=data, weights=weights, localizer=localizer, k_folds=2
+    )
```
znicholls, line 33: Test to be added right?

znicholls (commented):
Very nice

znicholls:
> The tests are going to be nightmare - not looking forward to that.

Looks quite a good start so far! I also recently watched [this](https://www.youtube.com/watch?v=3MNVP9-hglc), it requires quite a different coding style but it might be worth considering in future as testing becomes simpler.

mesmer/core/localized_covariance.py
znicholls: Ye not lovely but also gets the point across

mesmer/core/localized_covariance.py
znicholls: I liked it, made it much easier for me to understand what was going on

mesmer/core/localized_covariance.py
znicholls: I don't think so (as I think empirical is used to distinguish from true, but we don't have a true covariance so we don't care about that distinction)

mesmer/core/localized_covariance.py
znicholls: I like covariance as it's more explicit but cov also works

mesmer/core/localized_covariance.py
mathause: :+1:

mesmer/core/localized_covariance.py
mathause: I find it easier to read this way

mesmer/core/localized_covariance.py
mathause: :+1:

mesmer/core/localized_covariance.py
mathause: :+1:

tests/integration/test_localized_covariance.py
mathause: Yes, sorry - these are basically still working notes.

mesmer/core/localized_covariance.py
znicholls: Fair enough. I guess I find the other way easier because I don't have to think, "Does changing the order of multiplication of these matrices matter?" We've got tests though so doesn't matter much either way.

tests/integration/test_localized_covariance.py
znicholls: No worries

mathause:
Thanks for the feedback so far! Yes, once I realized I do not have to go via `calc_geodist_exact` and `gaspari_cohn` to calculate the "localizer" the tests were no longer quite so daunting.

Anyway, I think that's ready for another round of review - not a beauty but I tried :upside_down_face:

mesmer/core/utils.py
```diff
@@ -1,12 +1,74 @@
...
+    # ensure it's a list because we cannot get an item from an iterable
+    sequence = list(sequence)
+
+    for i, element in enumerate(sequence):
+
+        res = func(element, **kwargs)
+
+        if np.isneginf(res):
+            raise ValueError("`fun` returned `-inf`")
+        # skip element if inf is returned - not sure about this?
```
znicholls, line 54: Ye I would raise an error too...

znicholls (approved):
Looks like good steps!

mathause:
The failures are due to a setuptools update.

mesmer/core/utils.py
mathause: I now warn here - I use `np.inf` to indicate we got a singular matrix. I quickly checked `scipy.optimize.minimize` does not raise or warn for +/-inf.

mathause:
Ok, let's merge this. I am not so sure yet how I will make this work for xarray...

Thanks a ton for your reviews - @znicholls!

PR #186, opened 2022-08-04
black: ignore MESMER_M notebook

mathause:
@snath-xoc I am excluding your MESMER_M notebook from being reformated with black. I am not sure what your state is here but I don't want to mess this up for you. (However, we should eventually reformat your code but that can wait.)


mathause:
> O.K. no worries! What is black?

Good question - a code autoformatter: https://black.readthedocs.io/en/stable/

PR #188, opened 2022-08-12
use pyproject.toml [build-system] and setuptools_scm

mathause:
I did not plan to do this _now_ but as setuptool v64 causes problems with versioneer I went for it (instead of pinning setuptools). I do have a bit of experience with this setup by now, so I feel fairly confident this "just works". Maybe requires a re-install in your development environment.

@znicholls FYI.


mathause:
So much to "I know what I am doing" :joy:

mathause:
Ok falling back to `pkg_ressources` for python 3.7 - we cannot conditionally requires pkg_resources for python 3.7 in conda...

docs/source/conf.py
```diff
@@ -8,22 +8,24 @@
...
+    f"(c) 2021-{copyright_year} ETH Zurich (Land-climate dynamics group, Prof. S.I. "
+    "Seneviratne), MESMER contributors listed in AUTHORS"
 )
+
 authors = "Authors, see AUTHORS"
 author = authors
 
 # The short X.Y version
-version = get_versions()["version"].split("+")[0]
+version = mesmer.__version__.split("+")[0]
```
znicholls, line 26: In other packages I had to follow this pattern https://github.com/openscm/openscm-runner/blob/27eb7baddb686862c7996dbccabf34cea934fcb7/docs/source/conf.py#L19

I can't remember why exactly, but it was something to do with being installed or not being installed when building docs

.git_archival.txt
```diff
@@ -0,0 +1,4 @@
+node: $Format:%H$
```
znicholls, line 1: Out of curiosity , what is this file?

docs/source/conf.py
mathause: Oh - I wasn't aware that it's not (no longer?) installed per default in RTD. In xarray and regionmask we install it manually using pip (but I thought this is because it installs it with eager updates [making it impossible to pin packages]).

In any case RTD currently reports "mesmer 0.0.0 documentation" - I'll try your pattern to fix this in a new PR.

Related - for regionmask it reported a "dirty" version even for releases. This happens because RTD changes some files before installing the package. I used the following to work around this: regionmask/regionmask/pull/348

.git_archival.txt
mathause: I only just got this recently. If you define `<filename> export-subst` in a .gitattributes file, git will expand several placeholders when adding this file to an archive (from https://git-scm.com/docs/gitattributes). These placeholders can be the commit hash etc.

It currently still points to the wrong file (#189). With #189 merged:

```bash
mkdir archive
git archive HEAD > archive/HEAD.tar.gz
cd archive
tar -xvf HEAD.tar.gz
cat .git_archival.txt
```

returns 

```
node: d933fa6f0835203e8310ca68b7307ebc34e63c43
node-date: 2022-08-15T11:05:04+02:00
describe-name: v0.8.3-39-gd933fa6
ref-names: HEAD -> fix_gitattributes_versioneer, origin/fix_gitattributes_versioneer
```

I think `git archive` is used to create the files in https://github.com/MESMER-group/mesmer/releases/tag/v0.8.3. Versioneer does it slightly differntly by writing to `mesmer/_version.py export-subst`.

docs/source/conf.py
mathause: We currently do an install on RTD using "setuptools", i.e. `i.e., python ./setup.py install`, which is deprecated and may therefore not work together with pyproject.toml.

see #190 and

https://github.com/MESMER-group/mesmer/blob/d71bbe17e15f675cfdc3e2715f104a524e4a14a7/.readthedocs.yml#L10

PR #190, opened 2022-08-15
fix doc version on RTD and small cleanup

mathause:
(Should eventually) fix the wrong version number on RTD with the new pyproject.toml etc. installation.

Also cleans up some unused config in docs/conf.py.

---

We might need to go to a manual install - IIRC RTD does something along the lines `python -m pip install . --upgrade-strategy eager` which upgrades packages we pinned (well we did not pin any at the moment but still).

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,13 +1,15 @@
 version: 2
 build:
   os: ubuntu-20.04
+  jobs:
+    pre_install:
+        # see https://github.com/readthedocs/readthedocs.org/issues/8201
+      - git update-index --assume-unchanged docs/environment.yml docs/source/conf.py
+      # install regionmask, needs to be editable
+      - pip install -e .
   tools:
     python: mambaforge-4.10
 sphinx:
   configuration: docs/source/conf.py
-python:
-  install:
-  - method: pip
-    path: .
 conda:
   environment: docs/environment.yml
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -7,7 +7,7 @@ sphinx:
   configuration: docs/source/conf.py
 python:
   install:
-  - method: setuptools
+  - method: pip
     path: .
 conda:
   environment: docs/environment.yml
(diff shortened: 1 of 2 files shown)


mathause:
And indeed this does

```bash
python -m pip install --upgrade --upgrade-strategy eager --no-cache-dir .
```

and which returns

```
...
Successfully installed mesmer-emulator-0.8.3.post1.dev40+gb22796c.d20220815 numpy-1.23.2 pytz-2022.2.1 setuptools-65.0.0
```

1. it's a dirty install
2. it upgraded numpy and others

and we do not want this... Luckily I already figured out how to fix this for regionmask.

.readthedocs.yml
```diff
@@ -1,13 +1,15 @@
 version: 2
 build:
   os: ubuntu-20.04
+  jobs:
+    pre_install:
+        # see https://github.com/readthedocs/readthedocs.org/issues/8201
+      - git update-index --assume-unchanged docs/environment.yml docs/source/conf.py
+      # install regionmask, needs to be editable
+      - pip install -e .
```
mathause, line 9:
```suggestion
      - python -m pip install -e .
```

docs/source/conf.py
```diff
@@ -7,9 +7,30 @@
...
 import datetime
+from importlib.metadata import version
+import sys
+import subprocess
+import os
 
 import mesmer
 
+# -- Display version info ----------------------------------------------------
+#
```
mathause, line 18:
```suggestion
# for debugging on RTD
```

.readthedocs.yml
```diff
@@ -1,13 +1,15 @@
 version: 2
 build:
   os: ubuntu-20.04
+  jobs:
+    pre_install:
+        # see https://github.com/readthedocs/readthedocs.org/issues/8201
+      - git update-index --assume-unchanged docs/environment.yml docs/source/conf.py
+      # install regionmask, needs to be editable
```
mathause, line 8:
```suggestion
      # install mesmer, needs to be editable
```

mathause:
Ok, I think this now does what we want. It's always incredible how much this stuff takes even if you "know what you're doing".

PR #196, opened 2022-09-09
separate unit & integration tests

### LINKED ISSUE #120: Split tests into integration and regression tests
This will make it easier to understand what the point of the tests is

### LINKED ISSUE #148: test_calibrate_mesmer_and_draw_realisations.py is empty?
I just saw that [test_calibrate_mesmer_and_draw_realisations.py](https://github.com/MESMER-group/mesmer/blob/master/tests/integration/test_calibrate_mesmer_and_draw_realisations.py) is empty. @znicholls I assume that is not intentional?

mathause:
Separate unit & integration tests. I think most of the newer tests should be classified as unit tests and not as integration. Also moves the test `utils` to `mesmer.testing` (hmm should now add tests for them).

@znicholls I am not entirely sure this is what you had in mind in #120 - if not let me know (I may merge soonish, though but can update later)

 - [x] Closes #148, closes #120
 - [ ] Tests added


mathause:
> Also moves the test `utils` to `mesmer.testing` (hmm should now add tests for them).

opened #198 for this

znicholls (commented):
Yep exactly what I was thinking, thanks for pushing this along!

PR #202, opened 2022-09-12
mask_frac: rename, fix coords and pole behaviour

mathause:
- [x] Tests added

I found two issues when trying to use `mask_percentage` with xarray objects.

1. The coords were not exactly correct (due to numerical issues)
2. The fractional overlap at the poles was incorrect.

This PR fixes those two problems. Note that the problems were not relevant for mesmer so far! Also adds some basic tests.

I will double check with the author of regionmask if this makes sense.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -42,7 +42,7 @@ Breaking changes
 - Moved ``gaspari_cohn`` & ``calc_geodist_exact`` from ``io.load_constant_files`` to ``core.computation``
   (`#158 <https://github.com/MESMER-group/mesmer/issues/158>`_).
   By `Yann Quilcaille <https://github.com/yquilcaille>`_.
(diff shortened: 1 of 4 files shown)


CHANGELOG.rst
```diff
@@ -42,6 +42,9 @@ Breaking changes
 - Moved ``gaspari_cohn`` & ``calc_geodist_exact`` from ``io.load_constant_files`` to ``core.computation``
   (`#158 <https://github.com/MESMER-group/mesmer/issues/158>`_).
   By `Yann Quilcaille <https://github.com/yquilcaille>`_.
+ - The function `mask_percentage` has been renamed to :py:func:`utils.regionmaskcompat.mask_3D_frac_approx`
```
mathause, line 45:
```suggestion
 - The function ``mask_percentage`` has been renamed to :py:func:`utils.regionmaskcompat.mask_3D_frac_approx`
```

mathause:
> I will double check with the author of regionmask if this makes sense.

I had a quick chat with them and the change seems good to go!

PR #204, opened 2022-09-14
tests: move output to separate folder & allow saving params

mathause:
This PR does two things

- it moves the output test data in `tests/test-data/` to `tests/test-data/output/one_scen_one_ens`
- `pytest tests/integration/test_calibrate_mesmer.py --update-expected-files` now sets `save_params` to `True` and writes them to `tests/test-data/output/one_scen_one_ens/params`

The idea is to (1) allow writing a integration test with `two_scen_one_ens` eventually (2) write out the params so I can double check some values that are not in the "mesmer bundle". This is helpful to re-engineer the data flow without using Lea's code path.

cc @znicholls (mostly FYI I think)


znicholls (approved):
lgtm

Issue #209, opened 2022-10-24
add more calibration tests


mathause:
Currently we only test one single ensemble member for one scenario in 

https://github.com/MESMER-group/mesmer/blob/5a456657147b226d14211aba20f148c34fa881ce/tests/integration/test_calibrate_mesmer.py#L12

We should also add tests for one scenario two ens members and two scenarios. The potential disadvantage is that we lock in the current averaging (or maybe it's good, then we have to be explicit about changes).

PR #211, opened 2022-10-26
Test one scen multi ens

mathause:
- [x] Part of #209
 - [x] Tests added


mathause:
Ok, after all the preparation we finally have these tests.
- I went back to v0.8.3 to create `test-mesmer-bundle.pkl`
- there are some small fixes required that went wrong in the refactoring
- I may have to add a `--slow` pytest fixture (?) as the new test is not exactly fast...

PR #217, opened 2022-11-02
convert grids to/ from unstructured

### LINKED ISSUE #105: unstructured grid
This issue is relevant for  #65

---

Internally mesmer uses an unstructured grid. That is the lat and lon coords are not 2D but along a vector. When we start using xarray internally we need to name the non-coordinate dimension of the vector. @leabeusch suggests to use `"gp"` (for gridpoint). The most likely alternative would be `"cell"` (see details).

<details>

This is what some other models do:

- ICON: `cell`
- MPAS: `nCells`
- CAM with unstructured grid(?) (CAMSE?): `ncol`
- CLM (used internally only): `column` and `gridcell`
- AWI ocean model: `ncells`

</details>

Example:

```python
import xarray as xr
xr.set_options(display_style="text")

lat = [0.5, 0.5, 1.5, 1.5]
lon = [0.5, 1.5, 0.5, 1.5]
data = [0.5, 0.7, 0.8, 0.2]

ds = xr.Dataset(
    data_vars=dict(data=("gp", data)),
    coords={"lon": ("gp", lon), "lat": ("gp", lat)}
)
```

and the repr would look like
```
<xarray.DataArray 'data' (gp: 4)>
array([0.5, 0.7, 0.8, 0.2])
Coordinates:
    lon      (gp) float64 0.5 1.5 0.5 1.5
    lat      (gp) float64 0.5 0.5 1.5 1.5
Dimensions without coordinates: gp
```

- Note: I would not go for a `MultiIndex` because (i) it brings its own set of problems (ii) it should no longer be necessary after the index refactor of xarray (which is actually finally underway) (iii) we probably seldom need to select individual grid points from the 

- Obviously the array has more dimensions, likely `"time"` and `"member"` (or `"realization"`). 
<details>

```
Dimensions:  (member, time, gp)
Coordinates:
    time (time)
    lon  (gp)
    lat  (gp)
Dimensions without coordinates: member, gp
```
</details>

- xarray does not support two dimensions with the same name. Therefore we need new names for the geo distance matrix and the correlation matrix. We thought to subscript `"gp"`, `"lon"`, and `"lat"` with `"_i"`, and `"_j"`:

<details>

```python
geodist = xr.Dataset(
    data_vars=dict(
        dist=(("gp_j", "gp_i"), [
        [0.5, 0.7, 0.8, 0.2],
        [0.5, 0.7, 0.8, 0.2],
        [0.5, 0.7, 0.8, 0.2],
        [0.5, 0.7, 0.8, 0.2]
        ])
    ),
    coords={
        "lon_j": ("gp_j", lon),
        "lat_j": ("gp_j", lat),
        "lon_i": ("gp_i", lon),
        "lat_i": ("gp_i", lat)
    }
)
```
</details>

I.e. the array would look like:
```
<xarray.Dataset>
Dimensions:  (gp_i: 4, gp_j: 4)
Coordinates:
    lon_j    (gp_j) float64 0.5 ...
    lat_j    (gp_j) float64 0.5 ...
    lon_i    (gp_i) float64 0.5 ...
    lat_i    (gp_i) float64 0.5 ...
Dimensions without coordinates: gp_i, gp_j
Data variables:
    dist     (gp_j, gp_i) float64 0.5 ...
```

### LINKED ISSUE #117: round-trip land-only unstructured grid
I thought about how we can round-trip an unstructured land-only grid. We will in some way need the original grid. I am ignoring this for now but this could be stored as coords on the new dataset (but then all coords and dims need renaming...).

Related to #105

```python
import xarray as xr
from packaging.version import Version

def to_unstructured(ds, x_dim="lon", y_dim="lat", cell_dim="cell", multiindex=False):

    dims = {cell_dim: (y_dim, x_dim)}

    ds = ds.stack(dims)

    if not multiindex:
        # the behaviour changed in xarray 2022.06 (Index refactor)
        drop = True if Version(xr.__version__) > Version('2022.3') else False
        ds = ds.reset_index(list(dims.keys()), drop=drop)

    return ds


def to_landonly(
    ds, x_dim="lon", y_dim="lat", cell_dim="cell", x_coords="lon", y_coords="lat"
):

    mask = land_110.mask_3D(ds, lon_name=x_coords, lat_name=y_coords).squeeze(drop=True)

    # convert (lat, lon) -> (cells)
    ds = to_unstructured(ds, x_dim, y_dim, cell_dim)
    mask = to_unstructured(mask, x_dim, y_dim, cell_dim)

    return ds.sel({cell_dim: mask})


def from_unstructured(ds, ds_orig, x_dim="lon", y_dim="lat", cell_dim="cell"):

    if not isinstance(ds.indexes.get(cell_dim), pd.MultiIndex):
        ds = ds.set_index({cell_dim: (y_dim, x_dim)})

    ds = ds.unstack(cell_dim)

    return xr.align(ds, ds_orig, join="right")[0]
```

I don't comment the examples - have a look at the "Dimensions":

### Example 1: 1D lat and lon coords

```python
air = xr.tutorial.open_dataset("air_temperature")
air.attrs = {}

print("\n# Original dataset:\n")
print(air)

air_unstructured = to_unstructured(air)

print("\n# Unstructured:\n")
print(air_unstructured)

air_landonly = to_landonly(air)

print("\n# Land only:\n")
print(air_landonly)

air_restored = from_unstructured(air_landonly, air)

print("\n# Restored from land-only:\n")
print(air_restored)
```

```
Original dataset:
 - Dimensions:  (lat: 25, time: 2920, lon: 53)
Unstructured:
 - Dimensions:  (time: 2920, cell: 1325)
Land only:
 - Dimensions:  (time: 2920, cell: 517)
Restored from land-only:
 - Dimensions:  (lat: 25, time: 2920, lon: 53)
```

### Example 2: 2D lat and lon coords

```python
rasm = xr.tutorial.open_dataset("rasm")
# we require coords...
rasm = rasm.assign_coords(x=rasm.x, y=rasm.y)
rasm.attrs = {}

print("\n# Original dataset:\n")
print(rasm)

rasm_unstructured = to_unstructured(rasm, x_dim="x", y_dim="y")
print("\n# Unstructured:\n")
print(rasm_unstructured)

rasm_landonly = to_landonly(rasm, x_dim="x", y_dim="y", x_coords="xc", y_coords="yc")
print("\n# Land only:\n")
print(rasm_landonly)

rasm_restored = from_unstructured(rasm_landonly, rasm, x_dim="x", y_dim="y")

print("\n# Restored from land-only:\n")
print(rasm_restored)
```

```
Original dataset:
 - Dimensions:  (time: 36, y: 205, x: 275)
Unstructured:
 - Dimensions:  (time: 36, cell: 56375)
Land only:
 - Dimensions:  (time: 36, cell: 23472)
Restored from land-only:
 - Dimensions:  (x: 275, time: 36, y: 205)
```

mathause:
- [x] Closes #117, closes #105
 - [x] Tests added

---

This is the first of several of PRs which will add helper functions to deal with `Dataset` and `DataArray` objects. This first one deals with converting arrays from regular grids to unstructured grids (i.e. flattening the `lon` and `lat` dimension). What will follow are PRs on masking (land & antarctica), calculating the global mean, and potentially, stacking ensembles, and calculating anomalies.

I do have a number of questions - most of them concerning naming.

---

1. General question: The functions are currently in `mesmer.xarray_utils`. I imagined this as follows:

   ```python
   import mesmer.xarray_utils as mxu

   ds = mxu.to_unstructured(ds)
   ```

   Does this sound good? I had a number of other ideas but start to prefer this name - as the functionality is xarray-specific. However, I am open to suggestions.

   Other ideas I had:
   
   ```python
   mesmer.data_utils
   mesmer.preproc
   mesmer.process
   mesmer.tools
   mesmer.utils
   mesmer.xarray_utils
   ```

2. I now went with `cell_dim="cell"` - objections? See discussion in #105.

2. Should the function be (quasi-)top-level (1) or not (2)?

   ```python
   import mesmer.xarray_utils as mxu

   ds = mxu.to_unstructured(ds) # (1)
   ds = mxu.grid.to_unstructured(ds) # (2)
   ```

3. What about the name `unstructured`? I think it comes from models that have a truly unstructured grid (e.g. ICON), and their spatial data has to be 1D. Thus, the name may not be entirely correct in this [context](https://en.wikipedia.org/wiki/Types_of_mesh#Classification_of_grids) as in most cases we will be flattening a regular grid. So what is a better name? (I mean I am happy to keep the current name if people understand what it means.)

4. There is a difficulty with round-tripping because we remove `cells` that are `NaN`. We need the original coordinates to restore the original coordinates. I decided that the user will be responsible to supplying them. Other solutions did not convince me (#117). This may look like:

   ```python
   import mesmer.xarray_utils as mxu

   ds = mxu.grid.to_unstructured(ds) 
   coords_orig = ds.coords.to_dataset()[["lon", "lat"]]

   ds = mxu.grid.from_unstructured(ds, coords_orig) # (2)
   ```

This is a draft - I plan to tag some people later.

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -42,10 +42,10 @@ Data manipulation
 .. autosummary::
    :toctree: generated/
 
-   ~xarray_utils.stack_lat_lon
-   ~xarray_utils.unstack_lat_lon_and_align
-   ~xarray_utils.unstack_lat_lon
-   ~xarray_utils.align_to_coords
+   ~xarray_utils.grid.stack_lat_lon
+   ~xarray_utils.grid.unstack_lat_lon_and_align
+   ~xarray_utils.grid.unstack_lat_lon
+   ~xarray_utils.grid.align_to_coords
 
 Train mesmer
 ------------
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -36,6 +36,15 @@ Computation
    ~core.computation.calc_geodist_exact
    ~core.computation.gaspari_cohn
 
+Data manipulation
+-----------------
+
+.. autosummary::
+   :toctree: generated/
+
+   ~xarray_utils.to_unstructured
+   ~xarray_utils.from_unstructured
+
 Train mesmer
 ------------
 
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -42,8 +42,10 @@ Data manipulation
 .. autosummary::
    :toctree: generated/
 
-   ~xarray_utils.to_unstructured
-   ~xarray_utils.from_unstructured
+   ~xarray_utils.stack_lat_lon
+   ~xarray_utils.unstack_lat_lon_and_align
+   ~xarray_utils.unstack_lat_lon
+   ~xarray_utils.align_to_coords
 
 Train mesmer
 ------------
--- a/mesmer/xarray_utils/grid.py
+++ b/mesmer/xarray_utils/grid.py
@@ -4,7 +4,7 @@ from packaging.version import Version
 
 
 def to_unstructured(
-    obj, x_dim="lon", y_dim="lat", cell_dim="cell", multiindex=False, dropna=True
+    obj, *, x_dim="lon", y_dim="lat", cell_dim="cell", multiindex=False, dropna=True
 ):
     """stack a regular grid to an unstructured grid
 
@@ -47,7 +47,7 @@ def to_unstructured(
     return obj
 
 
-def from_unstructured(obj, coords_orig, x_dim="lon", y_dim="lat", cell_dim="cell"):
+def from_unstructured(obj, coords_orig, *, x_dim="lon", y_dim="lat", cell_dim="cell"):
     """unstack an unstructured grid to a regular grid
 
     Parameters
@@ -84,6 +84,6 @@ def from_unstructured(obj, coords_orig, x_dim="lon", y_dim="lat", cell_dim="cell
     obj = xr.align(obj, coords_orig, join="right")[0]
 
     # make sure non-dimension coords are correct
-    obj = obj.assign_coords(coords_orig)
+    obj = obj.assign_coords(coords_orig.coords)
 
     return obj
--- a/mesmer/xarray_utils/grid.py
+++ b/mesmer/xarray_utils/grid.py
@@ -30,14 +30,13 @@ def to_unstructured(
     """
 
     dims = {cell_dim: (y_dim, x_dim)}
-    # dims = {cell_dim: (x_dim, y_dim)}
 
     obj = obj.stack(dims)
 
     if not multiindex:
         # there is a bug in xarray v2022.06 (Index refactor)
-        if Version(xr.__version__) == Version("2022.3"):
-            raise TypeError("There is a bug in xarray v2022.03. Please update xarray.")
+        if Version(xr.__version__) == Version("2022.6"):
+            raise TypeError("There is a bug in xarray v2022.06. Please update xarray.")
 
         obj = obj.reset_index(cell_dim)
 
--- a/mesmer/xarray_utils/grid.py
+++ b/mesmer/xarray_utils/grid.py
@@ -1,0 +1,142 @@
+import pandas as pd
+import xarray as xr
+from packaging.version import Version
+
+
+def to_unstructured(
+    obj, x_dim="lon", y_dim="lat", cell_dim="cell", multiindex=False, dropna=True
+):
+    """stack a regular grid to an unstructured grid
+
+    Parameters
+    ----------
+    obj : xr.Dataset | xr.DataArray
+        Array to convert to an unstructured grid.
+    x_dim : str, default: "lon"
+        Name of the x-dimension.
+    y_dim : str, default: "lat"
+        Name of the y-dimension.
+    cell_dim : str, default: "cell"
+        Name of the new dimension.
+    multiindex : bool, default: False
+        If the new `cell_dim` should be returned as a MultiIndex.
+    dropna : bool, default: True
+        Drops each 'cell' if any NA values are present.
+
+    Returns
+    -------
+    obj : xr.Dataset | xr.DataArray
+        Array converted to an unstructured grid.
+    """
+
+    dims = {cell_dim: (y_dim, x_dim)}
+    # dims = {cell_dim: (x_dim, y_dim)}
+
+    obj = obj.stack(dims)
+
+    if not multiindex:
+        # there is a bug in xarray v2022.06 (Index refactor)
+        if Version(xr.__version__) == Version("2022.3"):
+            raise TypeError(
+                "There is a bug in xarray v2022.03." "Please update of xarray."
+            )
+
+        obj = obj.reset_index(cell_dim)
+
+    if dropna:
+        obj = obj.dropna(cell_dim)
+
+    return obj
+
+
+def from_unstructured(obj, coords_orig, x_dim="lon", y_dim="lat", cell_dim="cell"):
+    """unstack an unstructured grid to a regular grid
+
+    Parameters
+    ----------
+    obj : xr.Dataset | xr.DataArray
+        Array to convert to an unstructured grid.
+    coords_orig : xr.Dataset | xr.DataArray
+        xarray object containing the original coordinates before it was converted to the
+        unstructured grid.
+    x_dim : str, default: "lon"
+        Name of the x-dimension.
+    y_dim : str, default: "lat"
+        Name of the y-dimension.
+    cell_dim : str, default: "cell"
+        Name of the new dimension.
+    multiindex : bool, default: False
+        If the new `cell_dim` should be returned as a MultiIndex.
+    dropna : bool, default: True
+        Drops each 'cell' if any NA values are present.
+
+    Returns
+    -------
+    obj : xr.Dataset | xr.DataArray
+        Array converted to an unstructured grid.
+    """
+
+    # a MultiIndex is needed to unstack
+    if not isinstance(obj.indexes.get(cell_dim), pd.MultiIndex):
+        obj = obj.set_index({cell_dim: (y_dim, x_dim)})
+
+    obj = obj.unstack(cell_dim)
+
+    # ensure we don't loose entire rows/ columns
+    obj = xr.align(obj, coords_orig, join="right")[0]
+
+    # make sure non-dimension coords are correct
+    obj = obj.assign_coords(coords_orig)
+
+    return obj
+
+
+# import mesmer
+
+# tas = None
+# REFERENCE_PERIOD = None
+# x_dim = None
+# y_dim = None
+
+
+# tas = mesmer.tools.calc_anomaly(tas, reference_period=REFERENCE_PERIOD)
+
+# wgt = mesmer.tools.lat_weights(tas)
+# tas_globmean = mesmer.tools.calc_globmean(tas, wgt)
+
+
+# tas = mesmer.tools.mask_land(tas)
+# tas = mesmer.tools.mask_antarctica(tas)
+
+# coords_orig = tas[[x_dim, y_dim]]
+
+# tas = mesmer.tools.to_unstructured(tas)
+
+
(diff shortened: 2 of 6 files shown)


mesmer/xarray_utils/grid.py
```diff
@@ -0,0 +1,89 @@
...
+
+    dims = {cell_dim: (y_dim, x_dim)}
+    # dims = {cell_dim: (x_dim, y_dim)}
+
+    obj = obj.stack(dims)
+
+    if not multiindex:
+        # there is a bug in xarray v2022.06 (Index refactor)
+        if Version(xr.__version__) == Version("2022.3"):
+            raise TypeError("There is a bug in xarray v2022.03. Please update xarray.")
```
mathause, line 40:
```suggestion
        if Version(xr.__version__) == Version("2022.6"):
            raise TypeError("There is a bug in xarray v2022.06. Please update xarray.")
```

mesmer/xarray_utils/grid.py
```diff
@@ -0,0 +1,89 @@
...
+        Drops each 'cell' if any NA values are present.
+
+    Returns
+    -------
+    obj : xr.Dataset | xr.DataArray
+        Array converted to an unstructured grid.
+    """
+
+    dims = {cell_dim: (y_dim, x_dim)}
+    # dims = {cell_dim: (x_dim, y_dim)}
```
mathause, line 33:
```suggestion
```

znichollscr:
> 1. I had a number of other ideas but start to prefer this name - as the functionality is xarray-specific

If you like it, let's go with that

> 2\. objections?

None from me

> 3\. Should the function be (quasi-)top-level (1) or not (2)?

I would do 2 (users can always do `from mesmer.xarray_utils.grid import to_unstructured` if they want something else

> 4\. What about the name `unstructured`?

Hmm yes tricky. My instinct would be 'flat' or 'long' as they feel more like the usual data science terms. You could use the names `stack_lat_lon` and `unstack_lat_lon` so it's a bit clearer what is actually going on?

> 5\. There is a difficulty with round-tripping because we remove `cells` that are `NaN`

Do we have to do this? Maybe it's naive, but I feel like we should be able to preserve nans unless the user doesn't want to? If we have to keep the nans, then sounds fine (I'll make suggestions during review)

Thanks for a great PR!

mesmer/xarray_utils/grid.py
```diff
@@ -0,0 +1,88 @@
+import pandas as pd
+import xarray as xr
+from packaging.version import Version
+
+
+def to_unstructured(
+    obj, *, x_dim="lon", y_dim="lat", cell_dim="cell", multiindex=False, dropna=True
+):
+    """stack a regular grid to an unstructured grid
```
znichollscr, line 9:
```suggestion
    """Stack a lat-lon grid to a 1D grid
```

In line with comments about calling this 'stack' rather than 'to_unstructured' in an earlier comment

mesmer/xarray_utils/grid.py
```diff
@@ -0,0 +1,88 @@
...
+        Array converted to an unstructured grid.
+    """
+
+    # a MultiIndex is needed to unstack
+    if not isinstance(obj.indexes.get(cell_dim), pd.MultiIndex):
+        obj = obj.set_index({cell_dim: (y_dim, x_dim)})
+
+    obj = obj.unstack(cell_dim)
+
+    # ensure we don't loose entire rows/ columns
```
znichollscr, line 82: Would it make sense to pull these last two steps into a different function? So round tripping would look like one of the following

```python
start = xr.DataArray(...)

stacked = to_unstructured(start)

# by default, we just lost any missing rows, but that's fine
# question: do we only remove nans if dropna is True or 
# do some always get lost?
stacked_lost_nans = from_unstructured(stacked)

# if users want the nans, they can add them back in e.g.
stacked_incl_nans = align_to_coords(stacked_lost_nans, coords_orig)
```

mesmer/xarray_utils/grid.py
```diff
@@ -0,0 +1,88 @@
...
+    x_dim : str, default: "lon"
+        Name of the x-dimension.
+    y_dim : str, default: "lat"
+        Name of the y-dimension.
+    cell_dim : str, default: "cell"
+        Name of the new dimension.
+    multiindex : bool, default: False
+        If the new `cell_dim` should be returned as a MultiIndex.
+    dropna : bool, default: True
+        Drops each 'cell' if any NA values are present.
```
znichollscr, line 24: Does this drop cell by cell or somehow row by row?

znichollscr (commented):
lgtm, couple of questions about naming and exactly how things are setup

mathause (commented):
Thanks for the helpful feedback! I agree `stack` and `unstack` are more intuitive names (`flatten` might work as well but I would not know how to name the inverse operation). Let's see how this looks like:

```python
tas_masked = xru.mask.mask_ocean(tas)

tas_stacked = xru.grid.stack_lat_lon(tas_masked)

tas_unstacked = xr.grid.unstack_lat_lon(tas_stacked, coords_orig)
```

Maybe `cell_dim="cell"` should then be `stack_dim="cell"`?

---

We remove the `<NA>` because our statistical functions don't handle them, and, because we don't emulate the ocean (currently), it allows to remove ~70% of the data. That this cannot be round-tripped is an annoying little thing that bugs me without end. A minimal example:

<details>

```python
import numpy as np
import matplotlib.pyplot as plt
import xarray as xr

%matplotlib

air = xr.tutorial.open_dataset("air_temperature")
coords_orig = air[["lat", "lon"]]

air = air.air
# add a row of Na -> e.g. by masking the ocean
air[:, 12, :] = np.NaN

air_stacked = air.stack(cell=("lat", "lon"))
air_stacked = air_stacked.dropna("cell")

air_unstacked = air_stacked.unstack("cell")
air_unstacked_aligned = xr.align(air_unstacked, coords_orig, join="right")[0]

print(f"{air.lat.shape=}")
print(f"{air_unstacked.lat.shape=}")
print(f"{air_unstacked_aligned.lat.shape=}")

# ---

f, axs = plt.subplots(2, 1)
# difficult to see but one lat-band is too big
air_unstacked.isel(time=0).plot(ax=axs[0])
air_unstacked_aligned.isel(time=0).plot(ax=axs[1])
```

</details>

mesmer/xarray_utils/grid.py
znicholls: Worth adding that to the docs?

znicholls:
```suggestion
        Drops each 'cell' if any NA values are present at any point in the timeseries.
```

mesmer/xarray_utils/grid.py
znicholls: Yep looks good to me! `unstack_lat_lon_no_align` could be `unstack_lat_lon` and `unstack_lat_lon` could be `unstack_lat_lon_and_align` ?

znicholls: No strong preference there, just a thought...

znicholls:
> Maybe `cell_dim="cell"` should then be `stack_dim="cell"`?

Yep sounds good to me!

mathause (commented):
Ok I implemented the naming updates. I also changed `cell_dim="cell"` to `stack_dim="gridcell"`.

I'll merge in a few days unless I get more feedback.

docs/source/api.rst
```diff
@@ -36,6 +36,17 @@ Computation
...
+Data manipulation
+-----------------
+
+.. autosummary::
+   :toctree: generated/
+
+   ~xarray_utils.stack_lat_lon
+   ~xarray_utils.unstack_lat_lon_and_align
+   ~xarray_utils.unstack_lat_lon
+   ~xarray_utils.align_to_coords
```
mathause, line 48:
```suggestion
   ~xarray_utils.grid.stack_lat_lon
   ~xarray_utils.grid.unstack_lat_lon_and_align
   ~xarray_utils.grid.unstack_lat_lon
   ~xarray_utils.grid.align_to_coords
```

znicholls (approved):
Looks awesome, nice work!

mathause:
Thanks! My plan is to create a tutorial with a potential workflow once all these PRs are done.

PR #219, opened 2022-11-07
xarray_utils: mask ocean and Antarctica

mathause:
- [ ] Closes #xxx
 - [x] Tests added

The second one in the series of data manipulation PRs (after #217). This adds functions to mask out the ocean and Antarctica. I think this is quite a bit easier than #217.

The naming of `x_dim` vs. `x_coords` is intentional (compare in #217). For most cases this will not play a role and the two will be the same. However, for regional simulations with a rotated grid it will make a difference. I should probably do a terminology section as for [xarray](https://docs.xarray.dev/en/stable/user-guide/terminology.html).

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -42,8 +42,8 @@ Data manipulation
 .. autosummary::
    :toctree: generated/
 
-   ~xarray_utils.mask_land_fraction
-   ~xarray_utils.mask_land
+   ~xarray_utils.mask_ocean_fraction
+   ~xarray_utils.mask_ocean
    ~xarray_utils.mask_antarctica
 
 Train mesmer
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -36,6 +36,16 @@ Computation
    ~core.computation.calc_geodist_exact
    ~core.computation.gaspari_cohn
 
+Data manipulation
+-----------------
+
+.. autosummary::
+   :toctree: generated/
+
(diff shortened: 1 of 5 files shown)


mathause:
@jschwaab - if you have time I'd appreciate if you can take a look.

docs/source/api.rst
```diff
@@ -36,6 +36,16 @@ Computation
...
 
+Data manipulation
+-----------------
+
+.. autosummary::
+   :toctree: generated/
+
+   ~xarray_utils.mask_ocean_fraction
+   ~xarray_utils.mask_ocean
+   ~xarray_utils.mask_antarctica
```
mathause, line 47:
```suggestion
   ~xarray_utils.mask.mask_ocean_fraction
   ~xarray_utils.mask.mask_ocean
   ~xarray_utils.mask.mask_antarctica
```

jschwaab:
Looks great Mathias! The functions for masking out ocean and Antarctica are very clear now. My only suggestion would be to maybe include some sort of comment to make clear that the masking in `mask_land` (for irregular grids) is based on whether the center of the grid cell is over land or ocean.

I like the idea of keeping things simple using http://www.naturalearthdata.com `regionmask.defined_regions.natural_earth_v5_0_0.land_110` and not allowing for other masks

mathause:
@jschwaab Thanks a lot for your review. This is ready for a final look.

jschwaab (approved):
perfect!

PR #229, opened 2022-12-07
add test to train and emulate (legacy)

mathause:
- [ ] Closes #xxx
 - [ ] Tests added

Add the example (copied from `examples/train_create_emus_automated.py` and `configs/config_across_scen_T_cmip6ng_test.py`) as a test. Currently nothing is actually tested yet. I add this as WIP to see the influence on the coverage.


mathause:
I will _not_ turn this into a test that only uses tas as predictor - it's much easier with the `_calibrate_tas` function. However, I may still run the example (without actually checking the output).

tests/integration/test_train_and_emulate_legacy.py
```diff
@@ -0,0 +1,22 @@
+import importlib.resources
```
mathause, line 1: Only available in python 3.9+...

mathause:
> I think this can also be closed @mathause

Agreed (may still be worth to ensure the examples run)

PR #230, opened 2022-12-07
github actions: use micromamba

mathause:
Use micromamba instead of miniconda to create the test environment. That should cut down the time needed to setup the test environment considerably...


mathause:
Allright this brings the test time down to < 2 min from > 5 min. So I'd call this a win. What is interesting is that the cache hits seem to only make a difference of 15 s to 30 s (with a download size of 300 MB) which I find impressive...

mathause:
@znicholls FYI

PR #241, opened 2023-01-16
refactor concat_hist_future

### LINKED ISSUE #27: Get rid of concat_h_f argument in create_emus_lt() and create_emus_gt()
The task of concatenating the historical period and the future scenario should be carried out by a separate function instead of being part of the `create_emus` functions. This change is necessary because we eventually want to be able to create emulations for arbitrary emission scenarios, which will not necessarily contain a historical time period and a future scenario.

mathause:
- [ ] Closes  #27
 - [ ] Tests added

I am not sure this actually fixes #27, as it does not remove the `concat_h_f` argument. However, it does refactor the code and I think the logic is now easier to follow. @leabeusch what do you think?

### DIFF
--- a/mesmer/create_emulations/utils.py
+++ b/mesmer/create_emulations/utils.py
@@ -1,4 +1,54 @@
+import numpy as np
 import xarray as xr
+
+
+def _concatenate_hist_future(data):
+    """concatenate hist and future data
+
+    Parameters
+    ----------
+    data : dict
+        Possibly nested dictionary containing arrays to concatenate. The keys of data
+        must correspond to the scenarios to use. The values can either be numpy arrays
+        or dicts of numpy arrays.
+
+    Returns
+    -------
+    concatenated : dict
+        Possibly nested dictionary with concatenated arrays.
+    """
+
+    scens_in = list(data.keys())
+
+    if "hist" not in scens_in[0]:
+        raise ValueError("data does not contain 'hist' scenario")
+
+    scens_in.remove("hist")
+    scens_out = [f"h-{scen}" for scen in scens_in]
+
+    concatenated = {}
+
+    hist = data.pop("hist")
+
+    # data is a nested dict
+    if isinstance(hist, dict):
+
+        for scen_out, scen_in in zip(scens_out, scens_in):
+            concatenated[scen_out] = {}
+
+            for targ in data[scen_in].keys():
+                concatenated[scen_out][targ] = np.concatenate(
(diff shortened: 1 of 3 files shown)


mesmer/create_emulations/utils.py
```diff
@@ -1,6 +1,56 @@
...
+
+    Returns
+    -------
+    concatenated : dict
+        Possibly nested dictionary with concatenated arrays.
+    """
+
+    scens_in = list(data.keys())
+
+    if "hist" not in scens_in[0]:
```
mathause, line 23:
```suggestion
    if "hist" not in scens_in:
```

mesmer/create_emulations/utils.py
```diff
@@ -1,6 +1,56 @@
+import numpy as np
 import xarray as xr
 
 
+def _concatenate_hist_future(data):
+    """concatenate hist and future data
```
mathause, line 6:
```suggestion
    """concatenate historical and future data
```

mesmer/create_emulations/utils.py
```diff
@@ -1,6 +1,56 @@
...
+            for targ in data[scen_in].keys():
+                concatenated[scen_out][targ] = np.concatenate(
+                    [hist[targ], data[scen_in][targ]]
+                )
+
+    # data is not a nested dict
+    else:
+
+        for scen_out, scen_in in zip(scens_out, scens_in):
+
```
mathause, line 48:
```suggestion
```

mesmer/create_emulations/utils.py
```diff
@@ -1,6 +1,56 @@
...
+    concatenated = {}
+
+    hist = data.pop("hist")
+
+    # data is a nested dict
+    if isinstance(hist, dict):
+
+        for scen_out, scen_in in zip(scens_out, scens_in):
+            concatenated[scen_out] = {}
+
```
mathause, line 38:
```suggestion
```

mesmer/create_emulations/utils.py
```diff
@@ -1,6 +1,56 @@
...
+    scens_in.remove("hist")
+    scens_out = [f"h-{scen}" for scen in scens_in]
+
+    concatenated = {}
+
+    hist = data.pop("hist")
+
+    # data is a nested dict
+    if isinstance(hist, dict):
```
mathause, line 34: I don't like this if/ else but I don't see a good way to avoid it.

PR #251, opened 2023-04-05
pin pandas < 2.0 and black < 23.1

mathause:
pin black to a version below 23.1 to avoid collapsing the opening newline


mathause:
I think the failing tests are due to pandas 2.0 - so let's pin that as well...

PR #262, opened 2023-05-26
move v0 code to own folder

mathause:
This illustrates how we could move all code from v0 to it's own folder without disrupting the usage of the project. This is not something I want to do right now but wanted to have here for reference. Obviously the other folders would need to be moved as well.

Another option would be to keep `packages = find:`, add `legacy_mesmer/__init__.py`, add `from legacy_mesmer import calibrate_mesmer` to `mesmer/__init__.py`. _However_, `from mesmer.calibrate_mesmer import train_gt` will then fail.

 - [ ] Closes #xxx
 - [ ] Tests added


mathause:
This will happen after v0.9.0 and before v1.0.0 (have to decide if in v0.10.0 or in v0.11.0)

Issue #222, opened 2022-11-17
ship some data with mesmer?


mathause:
I though if we should officially ship some data with mesmer, e.g. observational datasets, example datasets, etc.

1. What data should be included?

   Currently I see two data sets that might be worth including (note: these are currently in the repository but not in the installed package):
   - aerosol optical thickness
   - example cmip6 data (for tests and tutorials)

   @MESMER-group/mesmer: are there other datasets that should be included?

2. How/ where should it be made available?
   There are several options how to ship the data, which will depend a bit on the amount of data... ((i) a `mesmer-data` repo, (ii) download from mesmer using [pooch](https://github.com/fatiando/pooch), (iii), package it into the installed mesmer package, ...)

cc @sarasita @snath-xoc

znicholls:
I've never done this properly, but my instinct would be either:

- a mesmer-data repo, with helper tools in mesmer to download that wrap around pooch
- put data for tests and tutorials in `tests`, put everything else in a mesmer-data repo or on zenodo

PR #267, opened 2023-07-11
provide aerosol data via pooch

mathause:
- [x] xref #222
 - [ ] Tests added

This has mesmer download and cache the aerosol data using pooch. I went for pooch because I have a bit of experience with it from regionmask. I think this approach should be fine, certainly as long as we stay in the MB range. pooch is added as additional dependency (however, regionmask also has pooch in its list of requirements so it's not really adding anything).

I expect the tests to fail because the data file is currently not yet available at the specified location on github. Maybe I do another PR for this.


znicholls:
Nice

mathause:
thanks!

PR #273, opened 2023-09-04
grid_wrap: roundtrip tests

mathause:
Note that the wrap functions do not strictly round trip data that is outside the -180..180/ 0..360 range.

 - [ ] Closes #xxx
 - [ ] Tests added


mathause:
I think the failed rtd build is spurious - merging.

PR #285, opened 2023-09-06
lowess: n_steps and use_coords

mathause:
- [ ] Closes #xxx
 - [ ] Tests added

This adds some changes to `mesmer.stats.smoothing.lowess`:
- add `n_steps` keyword, which can be used instead of `frac` (as this is used in mesmer)
- rename `use_coords_as_x` to `use_coords`
- set `use_coords=True` to use coords as x-values per default. For datetime coords this will have to be set to `False` - this is a bit unfortunate (because we will mostly use it for that). However, it's the right thing to do - we want to be explicit about not using the coords (important when they are not equally spaced).

### DIFF
--- a/tests/unit/test_smoothing.py
+++ b/tests/unit/test_smoothing.py
@@ -16,6 +16,29 @@ def test_lowess_errors():
     with pytest.raises(ValueError, match="data should be 1-dimensional"):
         mesmer.stats.smoothing.lowess(data.to_dataset(), "data", frac=0.3)
 
+    with pytest.raises(ValueError, match="Exactly one of ``n_steps`` and ``frac``"):
+        mesmer.stats.smoothing.lowess(data.to_dataset(), "lat")
+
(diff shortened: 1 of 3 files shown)


tests/unit/test_smoothing.py
```diff
@@ -16,6 +16,29 @@ def test_lowess_errors():
...
+
+    # numpy datetime
+    time = xr.date_range("2000-01-01", periods=30)
+    data = data.assign_coords(time=time)
+
+    with pytest.raises(TypeError, match="Cannot convert coords"):
+        mesmer.stats.smoothing.lowess(data.to_dataset(), "time", frac=0.5)
+
+    # cftime datetime
+    time = xr.date_range("2000-01-01", periods=30, calendar="noloeap")
```
mathause, line 36:
```suggestion
    time = xr.date_range("2000-01-01", periods=30, calendar="noleap")
```

PR #300, opened 2023-09-21
add calc_gaspari_cohn_correlation_matrices

mathause:
- [ ] Closes #xxx
 - [ ] Tests added

Add function to calculate Gaspari-Cohn correlation matrices for a range of localisation radii. Again, blatantly stolen from #109, so @znicholls should really get the credit here. I pass `geodist` instead of `lon, lat` because I still intend to do #62 and for this it is much easier if we pass the geodist directly. Makes it also a bit more explicit what is happening.

The tests are crap - they repeat the code in the function, which renders almost worthless. Still it checks that it is run.


mathause:
I'm ignoring the faulty rtd built. I think it's a fluke.

PR #331, opened 2023-11-13
lowess: allow to pool along dim

mathause:
- [ ] Closes #xxx
 - [x] Tests added

Add `combine_dim` argument to `lowess` so data can be pooled to estimate the lowess smoothing. This will probably never be used, but I only realized this once I finished implementing it...

### DIFF
--- a/mesmer/stats/smoothing.py
+++ b/mesmer/stats/smoothing.py
@@ -4,7 +4,9 @@ import xarray as xr
 from mesmer.core.utils import _check_dataarray_form
 
 
-def lowess(data, dim, *, n_steps=None, frac=None, use_coords=True, it=0):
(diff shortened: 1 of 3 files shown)


mesmer/stats/smoothing.py
```diff
@@ -76,6 +89,19 @@ def lowess(data, dim, *, n_steps=None, frac=None, use_coords=True, it=0):
         x = xr.ones_like(coords)
         x.data = np.arange(coords.size)
 
+    if combine_dim is not None:
+        # remove non-dimension coords along combine_dims
+
+        print(data[combine_dim].coords.keys())
+
```
mathause, line 96:
```suggestion
```

PR #361, opened 2023-12-19
lin reg: only warn on superfluous predictors

mathause:
- [ ] Closes #xxx
 - [ ] Tests added

Just had a use case for this. Not sure if a good idea or not.

### DIFF
--- a/mesmer/stats/_linear_regression.py
+++ b/mesmer/stats/_linear_regression.py
@@ -1,3 +1,4 @@
+import warnings
 from typing import Mapping, Optional
 
 import numpy as np
@@ -88,7 +89,7 @@ class LinearRegression:
         if available_predictors - required_predictors:
             superfluous = sorted(available_predictors - required_predictors)
             superfluous = "', '".join(superfluous)
-            raise ValueError(f"Superfluous predictors: '{superfluous}'")
+            warnings.warn(f"Superfluous predictors: '{superfluous}'")
 
         if "intercept" in exclude:
             prediction = xr.zeros_like(params.intercept)
--- a/tests/unit/test_linear_regression.py
+++ b/tests/unit/test_linear_regression.py
@@ -110,17 +110,25 @@ def test_lr_predict_missing_superfluous():
     )
     lr.params = params
 
+    da = xr.DataArray([0, 1, 2], dims="time")
+
     with pytest.raises(ValueError, match="Missing predictors: 'tas', 'tas2'"):
         lr.predict({})
 
     with pytest.raises(ValueError, match="Missing predictors: 'tas'"):
         lr.predict({"tas2": None})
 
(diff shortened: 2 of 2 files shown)


veni-vidi-vici-dormivi:
I would like to suggest taking into account `exclude`. So that you cannot only exclude predictor vars from the params but also from the predictors Dataset. This is because you might not want to change anything about the original predictor set but still only use part of them. This makes more sense when you DataTree or Dataset as a datastructure because then It might be less work to exclude a predictor from a Dataset instead of making two Datasets.

```python
def predict(
        self,
        predictors: dict[str, xr.DataArray] | DataTree | xr.Dataset,
        exclude=None,
    ) -> xr.DataArray:
        """
        Predict using the linear model.

        Parameters
        ----------
        predictors : dict of xr.DataArray | DataTree | xr.Dataset
            A dict of ``DataArray`` objects used as predictors or a ``DataTree``, holding each
            predictor in a leaf. Each predictor must be 1D and contain ``dim``. If predictors
            is a ``xr.Dataset``, it must have each predictor as a ``DataArray``.
        exclude : str or set of str, default: None
            Set of variables to exclude in the prediction. May include ``"intercept"``
            to initialize the prediction with 0.

        Returns
        -------
        prediction : xr.DataArray
            Returns predicted values.
        """

        params = self.params

        exclude = _to_set(exclude)

        non_predictor_vars = {"intercept", "weights", "fit_intercept"}
        required_predictors = set(params.data_vars) - non_predictor_vars - exclude
        available_predictors = set(predictors.keys()) - exclude

        if required_predictors - available_predictors:
            missing = sorted(required_predictors - available_predictors)
            missing = "', '".join(missing)
            raise ValueError(f"Missing predictors: '{missing}'")

        if available_predictors - required_predictors:
            superfluous = sorted(available_predictors - required_predictors)
            superfluous = "', '".join(superfluous)
            raise ValueError(
                f"Superfluous predictors: '{superfluous}', either params",
                "for this predictor are missing or you forgot to add it to 'exclude'.",
            )

        if "intercept" in exclude:
            prediction = xr.zeros_like(params.intercept)
        else:
            prediction = params.intercept

        # if predictors is a DataTree, rename all data variables to "pred" to avoid conflicts
        if isinstance(predictors, DataTree) and not predictors.equals(DataTree()):
            predictors = map_over_subtree(
                lambda ds: ds.rename({var: "pred" for var in ds.data_vars})
            )(predictors)

        for key in required_predictors:
            prediction = (predictors[key] * params[key]).transpose() + prediction

        prediction = (
            _extract_single_dataarray_from_dt(prediction)
            if isinstance(prediction, DataTree)
            else prediction
        )

        return prediction.rename("prediction")
```

veni-vidi-vici-dormivi:
I think we can close this @mathause

mathause:
> I think we can close this @mathause

Good for me thanks!

PR #365, opened 2024-01-03
drop python 3.7 support

### LINKED ISSUE #163: drop support for python 3.7
[NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html) says packages can drop support for py37 in December 2021. And most packages did (numpy, xarray, ...). So we can and should do this as well.

I label this as a good-first-issue because it does not include any code changes, but it's not trivial because we have to dig into how the infrastructure is set up.

mathause:
- [ ] Closes #163
 - [ ] Tests added


mathause:
python 3.7 is EOL so this overdue: https://devguide.python.org/versions/

PR #368, opened 2024-01-03
test on python 3.12

mathause:
- [ ] Closes #xxx
 - [ ] Tests added


mathause:
Interesting - pandas < 2 is not supported for py3.12 - so let's do that.

Issue #338, opened 2023-11-19
using `np.unique` to de-duplicate historical ensembles shuffles them


mathause:
In `separate_hist_future` the historical ensembles are de-duplicated using `np.unique`. This shuffles the ensemble members "randomly" because `np.unique` sorts the output!

**TLDR**: This messes up the order of the residual variability and underestimates its influence.

---

1. This is relevant for `train_gv`, as the target (`tas`) and one of the predictors (`gv_novolc_tas_s`, the residual variability) is split. Now the target and residual variability have random order, leading to a wrong result.
3. The "smooth" predictors (`tas`, `tas2`, `hfds`) are averaged over the ensemble members (i.e. are all the same for all ensemble members) - so for them it does not matter. 
4. This seems to have a small influence on the params of the forced response, but gives the residual variability more predictive power (unsurprisingly). I cannot check this for thoroughly but will add the example where I found it.
6. `separate_hist_future` is also used in `train_gt`, but that does not matter, because the hist simulations are then averaged-over.

---

The story is: I replicated a workflow with several ensemble members outside of "legacy" MESMER and the params of the "local forced response" did not line up. It took way too long comparing and trying to re-order the `targs` and `preds` of my and Lea's version. Only when I was able to rule out any error on my side, did it make click.

This was a bitch to figure out - I _know_ that `np.unique` sorts its output, but I never thought about its implications, when looking at `separate_hist_future`.

cc @leabeusch (FYI) and @znicholls who might find this interesting.

mathause:
And to complete this, an example:
```python
import numpy as np
arr = np.array([1, 0, 1, 5, 2])
np.sort(arr)
# array([0, 1, 2, 5])
```

And potential bugfix
```python
__, idx = np.unique(arr, return_index=True)
arr[np.sort(idx)]
# array([1, 0, 5, 2])
```

mathause:
This was fixed in #339 and released with v0.10.0

PR #369, opened 2024-01-03
require pandas v2

### LINKED ISSUE #353: update date parsing in read_csv for pandas v2
For pandas v 2 `read_csv` returns a warning: 

> Could not infer format, so each element will be parsed individually, falling back to `dateutil`. To ensure parsing is consistent and as-expected, please specify a format.

To avoid add `date_format={"year_month": "%Y %m"}`

https://github.com/MESMER-group/mesmer/blob/616c8f2d69b0ca0cd88e108711414d4633752ec4/mesmer/core/_data.py#L37

But only for pandas v2.0+

mathause:
- [x] Closes #353
 - [ ] Tests added

This changes the requirements from `pandas<2.0` to `pandas>=2.0`. The pickles saved with `joblib` which contain pandas v1 structures cannot be opened when pandas v2 is installed (and vice versa). This is generally bad and we need to switch to netCDFs (#65), _however_,  here I abuse this to disable opening pickles created with mesmer v0.9 and lower. Thus, pickles created containing #338 need mesmer version 0.9.

Some comments:
- the nice thing to do would be to allow backward compatibility with a warning. This is hindered by two things
  - we don't have a mesmer function to open the mesmer-bundle pickles, but we use vanilla `joblib.load`
  - there is no way (that I know of) to open a joblib pickles with pd v1  (
- NEP29 would require pandas v1.5 (i.e. one version before v2.0)


mathause:
Ok thing is that xarray v2023.04 is required to support pandas v2.0 (which makes sense as pandas v2.0 was released in April 2023). However, xarray v2023.04 requires python 3.9+. So I may have to drop python 3.8 here as well. It's a bit more aggressive than I'd like but I think I'll go for it.

PR #372, opened 2024-01-04
simplify and update doc environment

mathause:
There seems to be some bloat in the docs environment. Let's try to trim this down.

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -5,20 +5,12 @@ channels:
   - nodefaults
 
 dependencies:
-  - python=3.9
-  - black<23.1
+  - python=3.11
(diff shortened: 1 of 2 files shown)


mathause:
Curious, it increases the size of the environment - probably because I also switched to python 3.11. Let's test with the old python version.

docs/environment.yml
```diff
@@ -5,20 +5,12 @@ channels:
   - nodefaults
 
 dependencies:
-  - python=3.9
-  - black<23.1
+  - python=3.11
```
mathause, line 8:
```suggestion
  - python=3.9
```

PR #375, opened 2024-01-04
remove unused deploy-pypi workflow

mathause:
We now have a dedicated workflow uploading to pypi copied from xarray (see #373 and #374). Uses "Trusted Publisher" on pypi (see https://pypi.org/manage/project/mesmer-emulator/settings/publishing/). It uploads to pypi when creating a release with the tag name `"v.*"`.

FYI @znicholls


znicholls:
Yep nice. Lots of improvement in this space, copying xarray's set up sounds very nice

PR #387, opened 2024-02-02
Fix volcanic tests

### LINKED ISSUE #386: test volc influence: numerical issues
Although we regress the data against itself, there seem to be numerical problems for Mac, such that the regression coefficient is not _exactly_ 1. So we should replace the `assert_identical` with `assert_allclose`

https://github.com/MESMER-group/mesmer/blob/7667fe92a545111adf919cfd405f1a380dfc086b/tests/integration/test_volc.py#L42-L45

https://github.com/MESMER-group/mesmer/blob/7667fe92a545111adf919cfd405f1a380dfc086b/tests/integration/test_volc.py#L55

veni-vidi-vici-dormivi:
[x] Closes #386

### DIFF
--- a/tests/integration/test_volc.py
+++ b/tests/integration/test_volc.py
@@ -40,9 +40,9 @@ def test_fit_volcanic_influence_self():
     )
 
     result = mesmer.volc.fit_volcanic_influence(-aod, slice("1850", "2014"))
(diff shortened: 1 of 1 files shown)


mathause (approved):

mathause:
Great, thanks!

PR #391, opened 2024-02-07
add pre-commit config

### LINKED ISSUE #301: use pre-commit-ci?
To take care of our linting needs. See mathause/mplotutils#10 for an example. We will also need to allow "pre-commit ci" in the Settings.

We could then remove: 

https://github.com/MESMER-group/mesmer/blob/1358e63b09ddedcbff1e2d6259b69177733eef3f/.github/workflows/ci-cd-workflow.yml#L15

I think that would be a good first issue, with low-to-medium difficulty, as non-code stuff needs to be adapted.

veni-vidi-vici-dormivi:
Closes #301


mathause (approved):
Looks good - feel free to merge. (I changed the title a bit).

Issue #317, opened 2023-10-03
auto regression does not return the *co*variance


mathause:
Ok, this is a bit akward, but what I did in #309 was wrong - not in that it leads to a wrong result, but the name is wrong. The AR process returns the _variance_ and not the *co*variance!

What I found confusing is that once this variance is used to draw auto regression time series, and once the co-variance from another source is used. I should try to make that clearer.

mathause:
Ok all this confusion is self made :-( Originally there were two code paths for drawing AR samples

- Before refactoring:
  - uncorrelated samples -> used `np.random.normal` -> required `std` 
  - correlated samples -> used `np.random.multivariate_normal` -> required `cov` 
- After refactoring:
  - uncorrelated samples -> uses `np.random.multivariate_normal` -> requires 1D `cov` 
  - correlated samples -> uses `np.random.multivariate_normal` -> requires 2D `cov` 

I am very tempted to just revert and have `_fit_auto_regression_xr` return `std`. Especially because I am thinking about splitting the code path again (for performance reasons, see  #165).

 _However_, as we average the variability of the residuals (see #307) its better to return them as `var`. I think...

PR #395, opened 2024-02-12
Let fit AR return nobs

### LINKED ISSUE #319: have fit AR regression return nobs
AR regression should return the number of observations: `AR_result.nobs` (it's the same as `AR_result.resid.size`, i.e. the number of residuals that were used to estimate the variance). This is required to properly fix #307.

(However, as we average twice we need to think how we do this. Would we give too much weight to large ensembles?)

https://github.com/MESMER-group/mesmer/blob/ee3ea6873ed7b46dc969338d6d6544f73855abb0/mesmer/stats/auto_regression.py#L157

veni-vidi-vici-dormivi:
Closes #319 
Tests added
Documentation added and changed std to variance in _fit_auto_regression_np returns (see also #317).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,8 @@ v0.11.0 - unreleased
 
 New Features
 ^^^^^^^^^^^^
-- Added number of observations to the output of the AR process. #395
+- Added number of observations to the output of the AR process. `#395 <https://github.com/MESMER-group/mesmer/pull/395>`_.
+  By `Victoria Bauer`_.
 
 Breaking changes
 ^^^^^^^^^^^^^^^^
@@ -368,3 +369,4 @@ v0.8.0 - 2021-07-13
 .. _`Mathias Hauser`: https://github.com/mathause
 .. _`Yann Quilcaille`: https://github.com/yquilcaille
 .. _`Zeb Nicholls`: https://github.com/znicholls
+.. _`Victoria Bauer`: https://github.com/veni-vidi-vici-dormivi
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,7 @@ v0.11.0 - unreleased
 
 New Features
 ^^^^^^^^^^^^
+- Added number of observations to the output of the AR process. #395
 
 Breaking changes
 ^^^^^^^^^^^^^^^^
--- a/mesmer/stats/_auto_regression.py
+++ b/mesmer/stats/_auto_regression.py
@@ -511,20 +511,20 @@ def fit_auto_regression(data, dim, lags):
     -------
     :obj:`xr.Dataset`
         Dataset containing the estimated parameters of the ``intercept``, the AR
-        ``coeffs`` and the ``variance`` of the residuals.
+        ``coeffs``, the ``variance`` of the residuals and the number of observations ``nobs``.
     """
 
     if not isinstance(data, xr.DataArray):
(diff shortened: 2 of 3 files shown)


veni-vidi-vici-dormivi:
We could remove `old_names = False` in 
https://github.com/MESMER-group/mesmer/blob/4c95c5f834b5982b1d7c6a3293c27416dc03e051/mesmer/stats/_auto_regression.py#L567C4-L567C57
apparently it has been depreciated.

mesmer/stats/_auto_regression.py
```diff
@@ -511,20 +511,20 @@ def fit_auto_regression(data, dim, lags):
...
-    intercept, coeffs, variance = xr.apply_ufunc(
+    intercept, coeffs, variance, nobs = xr.apply_ufunc(
         _fit_auto_regression_np,
         data,
         input_core_dims=[[dim]],
-        output_core_dims=((), ("lags",), ()),
+        output_core_dims=((), ("lags",), (), ()),
         vectorize=True,
-        output_dtypes=[float, float, float],
+        output_dtypes=[float, float, float, float],
```
mathause, line 527: Should that be `int` instead of float?

tests/unit/test_auto_regression.py
```diff
@@ -456,14 +456,16 @@ def test_fit_auto_regression_xr_1D_values():
...
+    lags = 1
+    result = mesmer.stats.fit_auto_regression(data, "time", lags=lags)
 
     expected = xr.Dataset(
         {
             "intercept": 1.04728995,
             "coeffs": ("lags", [0.99682459]),
             "variance": 1.05381192,
             "lags": [1],
+            "nobs": len(data.values) - lags,
```
mathause, line 468:
```suggestion
            "nobs": data["time"].size - lags,
```

`len` works but I don't use it for arrays - I prefer `arr.shape[axis]` (or `ds.sizes["time"]` for xarray objects).

(or `arr.size` if it's 1D (the total number of elements on the array))

mathause (commented):
Looks good. Can you add a small entry in the changelog?

> forgotten entry from https://github.com/MESMER-group/mesmer/issues/317.

What's the forgotten entry?

> We could remove `old_names = False` in 
https://github.com/MESMER-group/mesmer/blob/4c95c5f834b5982b1d7c6a3293c27416dc03e051/mesmer/stats/_auto_regression.py#L567C4-L567C57
apparently it has been depreciated.

It's currently a bit difficult, as we don't set the minimum defined versions of our dependencies. I think this only became relevant in statsmodels v0.14 which is still the newest version at the moment. So yes we'll have to eventually do this but maybe not here.

tests/unit/test_auto_regression.py
veni-vidi-vici-dormivi: > What's the forgotten entry?
In _fit_auto_regression_np the documentation still said it returns std instead of variance, I adapted the description.

> So yes we'll have to eventually do this but maybe not here.
Ok.

CHANGELOG.rst
```diff
@@ -6,6 +6,7 @@ v0.11.0 - unreleased
 
 New Features
 ^^^^^^^^^^^^
+- Added number of observations to the output of the AR process. #395
```
mathause, line 9: Can you add the full link to the PR and your name - check the other entries (you will have to add a link to your name at the end of the document).

mathause (approved):
LGTM - thanks. The changelog is formatted a bit differently.

PR #405, opened 2024-02-27
Remove old_names argument

### LINKED ISSUE #396: remove `old_names` argument
We use the `old_names` argument from statsmodels:

https://github.com/MESMER-group/mesmer/blob/3a6a9e154458cd7b93f928eda2fa3567ac902f53/mesmer/stats/_auto_regression.py#L570

and in the tests

https://github.com/MESMER-group/mesmer/blob/3a6a9e154458cd7b93f928eda2fa3567ac902f53/tests/unit/test_auto_regression.py#L567

we should remove this and require statsmodels v0.13+ (see #369 how to do this)

---

statsmodels is the first that has the default of `old_names=False`. See

- v0.12: https://github.com/statsmodels/statsmodels/blob/maintenance/0.12.x/statsmodels/tsa/ar_model.py#L254
- v0.13: https://github.com/statsmodels/statsmodels/blob/maintenance/0.13.x/statsmodels/tsa/ar_model.py#L192

---

I am not sure if statsmodels will deprecate the "old names" only or if it also deprecates the `old_names` argument. The reason why it does not raise a warning is

https://github.com/statsmodels/statsmodels/blob/37bd2e421bad61bc54e8260b86a10a68d17892b3/statsmodels/tsa/ar_model.py#L251

(They should set `old_names=None` and do

```python
if old_names is not None:
    warnings.warn(...)
```

)

veni-vidi-vici-dormivi:
- [x] Closes #396 


veni-vidi-vici-dormivi:
Note to self: we don't need any depreciation wrangling here because the old_names argument is only used internally, never exposed to the user.

veni-vidi-vici-dormivi:
Do we want a CHANGELOG entry for things like this, given it is an exclusively internal change. Or maybe is it common to have a CHANGELOG entry saying internal adjustments to dependency changes in PRs a b and c... ?

mathause:
> Do we want a CHANGELOG entry for things like this @mathause, given it is an exclusively internal change. Or maybe is it common to have a CHANGELOG entry saying internal adjustments to dependency changes in PRs a b and c... ?

We don't need a changelog entry for every PR & I decide on a case by case basis; for strictly internal changes I often don't add one. Here I'd add statsmodels to the list of "supported versions".

Can you add the statsmodels pin to setup.cfg?

mathause:
b.t.w. its deprecation (only one i)

CHANGELOG.rst
```diff
@@ -19,6 +19,8 @@ Breaking changes
   ============ ============= =========
   regionmask    >=0.8        >=0.9
   ============ ============= =========
+  statsmodels   any          >=0.13
```
mathause, line 22:
```suggestion
  statsmodels  not specified >=0.13
```

(`any` is fine - just to be consistent with below)

CHANGELOG.rst
```diff
@@ -19,6 +19,8 @@ Breaking changes
   ============ ============= =========
   regionmask    >=0.8        >=0.9
   ============ ============= =========
```
mathause, line 21:
```suggestion
```

mathause (commented):
Feel free to add the PR number above the list

mathause (approved):
Thanks

PR #406, opened 2024-02-28
Add warning if first element is selected as local min

### LINKED ISSUE #403: geodist selection: warn if the first element is selected
`find_localized_empirical_covariance` warns if it selects the last element (i.e. it finds no local minimum). 

```
/home/mathause/code/mesmer/mesmer/core/utils.py:71: OptimizeWarning: No local minimum found, returning the last element
  warnings.warn("No local minimum found, returning the last element", OptimizeWarning)
```

However, it does not warn if the first element is selected - this could imply that the minimum is before the selected one (e.g. if `LOCALISATION_RADII=range(10_000, 15_000, 1000)`)

```python
import mesmer

# warns
mesmer.core.utils._minimize_local_discrete(lambda x: [3, 2, 1][x], [0, 1, 2])
# does NOT warn
mesmer.core.utils._minimize_local_discrete(lambda x: [1, 2, 3][x], [0, 1, 2])
```

veni-vidi-vici-dormivi:
- [x] Closes #403 
 - [x] Tests added

### DIFF
--- a/mesmer/core/utils.py
+++ b/mesmer/core/utils.py
@@ -66,7 +66,10 @@ def _minimize_local_discrete(func, sequence, **kwargs):
         if res < current_min:
             current_min = res
         else:
-            return sequence[i - 1]
+            min_element = sequence[i - 1]
+            if min_element == sequence[0]:
+                warnings.warn("First element is local minimum.", OptimizeWarning)
+            return min_element
 
     warnings.warn("No local minimum found, returning the last element", OptimizeWarning)
 
--- a/mesmer/core/utils.py
+++ b/mesmer/core/utils.py
@@ -66,10 +66,11 @@ def _minimize_local_discrete(func, sequence, **kwargs):
         if res < current_min:
             current_min = res
         else:
-            min_element = sequence[i - 1]
-            if min_element == sequence[0]:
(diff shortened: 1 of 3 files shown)


mesmer/core/utils.py
```diff
@@ -66,7 +66,10 @@ def _minimize_local_discrete(func, sequence, **kwargs):
         if res < current_min:
             current_min = res
         else:
-            return sequence[i - 1]
+            min_element = sequence[i - 1]
+            if min_element == sequence[0]:
+                warnings.warn("First element is local minimum.", OptimizeWarning)
+            return min_element
```
mathause, line 72:
```suggestion
            # need to return element from the previous iteration
            sel = i - 1
            if sel == 0:
                warnings.warn("First element is local minimum.", OptimizeWarning)
            return sequence[sel]
```

mathause (commented):
Whichever approach you think is easier.

mesmer/core/utils.py
veni-vidi-vici-dormivi: Thanks, yours is even easier to understand I think.

mathause (approved):

PR #408, opened 2024-02-28
Test if drwaing realizations is close

### LINKED ISSUE #402: why are emulations not stable?
There is a test where we only check that 90% of the data is equal - this is ugly:

https://github.com/MESMER-group/mesmer/blob/00cabf9d5b56a5d970dff16f0bd4028f02f53ad6/tests/integration/test_draw_realisations_from_bundle.py#L105

Would be interesting to get to the bottom of this. However, could be nontrivial to figure out. My suspicion is that it comes from the matrix inversion.

- un-comment `assert_allclose` and check if this passes locally

   https://github.com/MESMER-group/mesmer/blob/00cabf9d5b56a5d970dff16f0bd4028f02f53ad6/tests/integration/test_draw_realisations_from_bundle.py#L113
- Do the same on GHA
- Potentially run on different OS on GHA (mac, win linux)
- Write a function to invert the matrices and check if this function leads to different results (see https://github.com/MESMER-group/mesmer/issues/168#issuecomment-1847271872)
- If we can identify `svd` as the culprit, this would be an additional motivation to save the 'inverted' covariance matrix (in addition to the speed gain, as it only needs to be inverted once).

veni-vidi-vici-dormivi:
Closes #402

### DIFF
--- a/tests/integration/test_draw_realisations_from_bundle.py
+++ b/tests/integration/test_draw_realisations_from_bundle.py
@@ -110,7 +110,7 @@ def test_make_realisations(
 
         # # Ideally we would use the below, but we can't because of numpy's
         # # random seed issue (see comment above).
-        # xr.testing.assert_allclose(result, exp, rtol=rtol)
+        xr.testing.assert_allclose(result, exp)
 
         # make sure we can get onto a lat lon grid from what is saved
         exp_reshaped = exp.set_index(z=("lat", "lon")).unstack("z")


veni-vidi-vici-dormivi:
So this looks good. I can remove the `_assert_frac_close` bit then. Do we need to update the test data too or is this done via GitHub?

mathause:
Interesting :wink: can you merge main (after #409). If you add `${{ matrix.os }}` to the name on L16 we'll directly see which OS fails.

veni-vidi-vici-dormivi:
☹️

veni-vidi-vici-dormivi:
I wanted to test with a new dataset but can't save a new one, see: https://github.com/pydata/xarray/issues/8743 .

veni-vidi-vici-dormivi:
Well now I am really confused 😅

veni-vidi-vici-dormivi:
macOS-latest is macOS 12 and I created the dataset on macOS 14 so let's try that.

veni-vidi-vici-dormivi:
Well would you look at that. The dataset from macos 14 succeeds on macOS 14 and windows latest.

mathause:
> Well would you look at that. The dataset from macos 14 succeeds on macOS 14 and windows latest.

Interesting that `macOS-latest` does not seem to be the latest macOS :man_shrugging:

veni-vidi-vici-dormivi:
Yes, but they are open about it: 
From https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources:
Note: The -latest runner images are the latest stable images that GitHub provides, and might not be the most recent version of the operating system available from the operating system vendor.

veni-vidi-vici-dormivi:
It bugs me that I cannot reproduce the large difference of the datasets we get between ubuntu and macOS. The test passed locally when I test the macOS dataset on atmos (openSUSE). GitHub Actions doesn't provide openSuse but one can set up self hosted runners, see [here](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/adding-self-hosted-runners). However, this is not recommended for public repositories for obvious reasons. Do we have an ubuntu machine where we could test where things are trailing off? 

Here are some observations I have made:
- The test passes on Mac and atmos locally with data generated on Mac
-  Between atmos and Mac I find differences for all emulation steps:
    - maximum differences in the global variability emulations are on the order of e-16
    - maximum differences in the local trend response are on the order of e-15
    - maximum differences in the local variability are on the order of e-13
    - maximum differences in the resulting emulations are on the order of e-13
- looking at the difference between the inverted matrices on the two systems (using numpy.linalg.inv and scipy.linalg.inv) I find maximum differences of e-14. The maximum difference between the numpy and scipy methods is also e-14.
- The params of all emulation steps are equal
- The fact we see small differences already for the global variability points to the fact that it is something about the random number generator as well not just the matrix inversion. Interestingly however, is that the differences in global variability (and subsequently local trend emulations) are just happening sometimes, the majority of the array is exactly equal. For the inverted matrices, the local variability emulations and the result, the whole array is slightly off.

All of this did not get me much closer to understand why the difference between the result on ubuntu and Mac is up to 0.8.

mathause:
I tried some more stuff:

- on cfc and on my laptop I get large differences
- on argon I do not
- on argon and cfc I use the same environment

I checked `np.show_config()`. What I saw it that argon has more simd extensions, which are maybe used for matrix inversions. But this is a wild guess at the moment.

<details>

``
  - AVX512F
  - AVX512CD
  - AVX512_SKX
  - AVX512_CLX
  - AVX512_CNL
  - AVX512_ICL
```
</details>

**Comment by veni-vidi-vici-dormivi** [comment]
Okay but that's something. Then I'll try to pin it down with output from cfc. Thanks!

**Comment by mathause** [comment]
In https://github.com/MESMER-group/mesmer/issues/168#issuecomment-1847337357 I found that small differences (1e-15), can lead to **one** large difference for the "inverted" matrices (0.1), which lead to differences ind the samples.

Yes, using cfc can definitely get you faster feedback.

**Comment by veni-vidi-vici-dormivi** [comment]
Okay, saving A that is used to draw from a multivariate makes the emulations pass on different OS. However, this just shifts our problem to testing the calibration where we save A. Now this will not be the same between different OS but one could argue this is less important to the user since the calibration can always be exported. 

A second issue here is that I was not actually able to reproduce the behavior of `numpy.random.multivariate_normal`, even though I think I copied the [exact Code](https://github.com/numpy/numpy/blob/main/numpy/random/mtrand.pyx#L4080) they use, but maybe I am wrong?

**Comment by veni-vidi-vici-dormivi** [comment]
Uff sorry the commits are a mess here, actually "fix dimensions of innovations" also includes switching from the numpy multivariate normal to the scipy multivariate normal.

**Comment by veni-vidi-vici-dormivi** [comment]
So this looks pretty good! I see two thing to do now:
1) test this approach with actual emulations and see if they are plausible
2) also implement cholesky in the localized covariance module, but maybe in another PR? but then we would want to check again the emulations...

File: tests/integration/test_draw_realisations_from_bundle.py [hunk 1]
```diff
@@ -110,7 +110,12 @@ def test_make_realisations(
 
         # # Ideally we would use the below, but we can't because of numpy's
         # # random seed issue (see comment above).
-        # xr.testing.assert_allclose(result, exp, rtol=rtol)
+        try:
+            xr.testing.assert_allclose(result, exp)
+        except AssertionError as e:
+            diff = (result - exp).to_array().values
+            print("\n max difference between old and new:", np.max(np.abs(diff)))
```
mathause, line 117: Remove this? Or else maybe use `warnings.warn(...)` (so it's at the end of the output)

mesmer/stats/_auto_regression.py
```diff
@@ -479,11 +482,23 @@ def _draw_auto_regression_correlated_np(
...
     # NOTE: 'innovations' is the error or noise term.
     # innovations has shape (n_samples, n_ts + buffer, n_coeffs)
-    innovations = np.random.multivariate_normal(
+    try:
+        cov = scipy.stats.Covariance.from_cholesky(np.linalg.cholesky(covariance))
+    except np.linalg.LinAlgError as e:
+        if "Matrix is not positive definite" in str(e):
+            w, v = np.linalg.eigh(covariance)  # , driver = "evr")
+            cov = scipy.stats.Covariance.from_eigendecomposition((w, v))
+            warnings.warn(
```
mathause, line 491: Use a `LinalgWarning` here? (`from mesmer.core.utils import LinAlgWarning`)

mathause (approved):
Very nice! Can you
- rename the PR
- open a issue that we re-consider the `try ... except np.linalg.LinAlgError` approach? (potentially require `method="eigh"`)


> 1. test this approach with actual emulations and see if they are plausible
> 2. also implement cholesky in the localized covariance module, but maybe in another PR? but then we would want to check again the emulations...

Yes and yes. Maybe do 1. after 2.?

tests/integration/test_draw_realisations_from_bundle.py
veni-vidi-vici-dormivi: Ah yes this still needs cleanup. We could now also remove the 90% test, are you okay with that?

mathause: sure!

mathause (approved):
~I'll assume the changelog entry is for the coming PR?~

edit: sorry my bad, all good

mathause:
Hurray 😄 Did you test the training and the drawing?

veni-vidi-vici-dormivi:
> Did you test the training and the drawing?

Yes, so calibrating + emulating together is 10 times faster. Drawing from the autocorrelated process is 10 times faster and finding the localized covariance is 22 times faster. `find_localized_empirical_covariance` takes the most time in both cases.

mathause:
awesome!

PR #409, opened 2024-02-28
GHA: don't fail fast

mathause:
Directly relevant for #408 but probably generally a good idea here.


veni-vidi-vici-dormivi:
Thanks!

PR #415, opened 2024-03-20
Integrate MESMER-M harmonic model

veni-vidi-vici-dormivi:
rewriting and fixing the harmonic model script of MESMER-M


veni-vidi-vici-dormivi:
@mathause I'm not sure if I want to keep working on this (work on all the todos I made in the harmonic model script) now or if I want to merge this and work on the power transformer in a new PR and then go back and optimize and clean up both together. What do you think?

mesmer/mesmer_m/utils.py
mathause, line 1: Can you add this in a separate PR?


- Not sure if it's the best location, but for now maybe add to
   https://github.com/MESMER-group/mesmer/blob/79c8ee01317528a54e05a330b25c11254105aac8/mesmer/core/utils.py#L93
- You may have to infer the fill direction using `xr.infer_freq`
- Would `annual_data.reindex_like(monthly_data)` work?
- (need to test what happens if date is in the middle of the month)

mesmer/mesmer_m/harmonic_model.py
```diff
@@ -0,0 +1,238 @@
...
+    predictions: array-like of shape (n_samples,)
+        Fourier Series of order n calculated over x and mon with coeffs.
+
+    """
+    # TODO: infer order from coeffs, rename n to order
+
+    #set these parameters fixed according to paper
+    # we could also fit them and give an inital guess of 0 and 1 in the coeffs array as before
+    beta0 = 0
+    beta1 = 1
```
mathause, line 42: Is that really constant? (That would be weird...)

mathause (commented):
> @mathause I'm not sure if I want to keep working on this (work on all the todos I made in the harmonic model script) now or if I want to merge this and work on the power transformer in a new PR and then go back and optimize and clean up both together. What do you think?

I think cleaning will happen in several PRs in any case. So I am fine to merge in the current state.

Two points to consider:
- It could be easier to see what you did, if you leave it in the mesmer_m folder. I leave it up to you if that is too cumbersome.
- Also keeping `tas_emu_monthly_full_eval-v1.ipynb` could be helpful if we want to have a look at my laptop.

mesmer/mesmer_m/harmonic_model.py
veni-vidi-vici-dormivi: So in the paper it says it's constant. In the Code it was fit before with initial guesses of 0 and 1 and the results were always very close to 0 and 1. Would you say we should fit them?

mathause: Ahh yes - it's the **local** annual mean temperature :facepalm: So it's 1 "per definition". However, if this is the case we should probably calculate anomalies and don't have them in the fitting routine. I'll open an issue.

veni-vidi-vici-dormivi: Thanks!

mathause:
About `months % 12 + 1`. This would be correct if months were enumerated as `0...n`. However, for months `1, .., 12, 1, ..., 12` it is likely wrong.

```python
import numpy as np

print(np.arange(0, 12) % 12 + 1)
print(np.arange(1, 13) % 12 + 1)
```

mesmer/mesmer_m/harmonic_model.py
```diff
@@ -152,65 +135,71 @@ def calculate_bic(n_samples, n_order, mse):
...
-    coeffs_fit, y_pred = fit_fourier_series_np(x, y, n_sel, repeat=repeat)
+    selected_order = np.argmin(bic_score) + 1
+    # TODO: we already fit for this order above so it would probably be faster to save the output from above and return it?
+    coeffs_fit, predictions = fit_fourier_series_np(yearly_predictor = yearly_predictor, 
+                                              monthly_target = monthly_target, 
+                                              order = selected_order)
 
-    coeffs = np.zeros([max_n * 4 - 2])
-    coeffs[: len(coeffs_fit)] = coeffs_fit
+    coeffs = np.zeros([max_order * 4]) # removed -2, because we always return all coefficients
```
mathause, line 184: Is `0` or `nan` better here? (Can also change later).

mesmer/mesmer_m/harmonic_model.py
```diff
@@ -152,65 +135,71 @@ def calculate_bic(n_samples, n_order, mse):
...
+    coeffs[: len(coeffs_fit)] = coeffs_fit # need the coeff array to be the same size for all orders
```
mathause, line 185:
```suggestion
    coeffs[: selected_order] = coeffs_fit # need the coeff array to be the same size for all orders
```

mathause (commented):
We still need to include mesmer_m in the pre-commit hook.

https://github.com/MESMER-group/mesmer/blob/f0bfc782475fcaeb8f7b62342464478bd8b3e656/.pre-commit-config.yaml#L4

We should do it in a separate PR, but I am not sure if _before_ or _after_ this one.

mesmer/mesmer_m/harmonic_model.py
```diff
@@ -76,72 +82,49 @@ def fit_fourier_series_np(x, y, n, repeat=False):
...
-    #            ).T
-    #            for i_n in range(n)
-    #        ]
-    #    )
-    # )
-
-    def fun(x, n, x_train, mon_train, y):
+    #mon_train = (np.pi * (mon_train % 12 + 1)) / 6 # this is double if we have it in generate_fourier_series_np as well
+
+    def fun(coeffs, n, x_train, mon_train, mon_target):
```
mathause, line 98:
```suggestion
    def func(coeffs, n, x_train, mon_train, mon_target):
```

I think it's more common to call it func

mesmer/__init__.py
```diff
@@ -42,6 +42,7 @@
     "testing",
     "volc",
     "weighted",
+    "mesmer_m",
```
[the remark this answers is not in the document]
veni-vidi-vici-dormivi: Wow thanks, I missed that. Probably happened when I merged the prep in.

mesmer/mesmer_m/harmonic_model.py
```diff
@@ -76,72 +82,49 @@ def fit_fourier_series_np(x, y, n, repeat=False):
...
     # NOTE: this seems to select less 'orders' than the scipy one
     # np.linalg.lstsq(A, y)[0]
 
     coeffs = optimize.least_squares(
-        fun, x0, args=(n, x_train, mon_train, y), loss="cauchy"
+        fun, firstguess, args=(order, x_train, mon_train, monthly_target), loss="cauchy"
     ).x
 
-    y_pred = generate_fourier_series_np(coeffs, n, x_train, mon_train)
+    preds = generate_fourier_series_np(coeffs = coeffs, order = order, yearly_T = x_train, months = mon_train)
```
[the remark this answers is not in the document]
veni-vidi-vici-dormivi: This is actually funny, in R you are supposed to always put the spaces for the default linter so I still do it sometimes 😅

veni-vidi-vici-dormivi:
Hm before would be prettier but the merge conflicts would probably be easier to handle after... Although this might not be true since I now already made the pull request (to see what happens) see #421. 😅

mesmer/mesmer_m/harmonic_model.py
```diff
@@ -76,72 +83,57 @@ def fit_fourier_series_np(x, y, n, repeat=False):
...
+            ** 2
+        )
 
         return loss
 
-    # print(A.shape,)
-    x0 = np.zeros(n * 4)
-    x0[2] = 1
-    x0[3] = 0
+    first_guess = np.zeros(order * 4)
```
mathause, line 108: Not for now but could start with the result from `order - 1`

mesmer/mesmer_m/harmonic_model.py
```diff
@@ -11,56 +11,63 @@
 import xarray as xr
 from scipy import optimize
 
+import mesmer.core.utils
```
mathause, line 14: Should make sure we don't need to import sub modules (i.e. `mesmer.core.utils` is available after `import mesmer`)

mesmer/mesmer_m/harmonic_model.py
```diff
@@ -152,65 +144,79 @@ def calculate_bic(n_samples, n_order, mse):
 
     """
 
-    n_params = n_order * 4 - 2
+    n_params = (
+        order * 4
+    )  # -2 because for order 0 the first two coefficients dissapear for sin(0)= 0
+    # removed this now because we don+t fit the beta0 and beta1
```
mathause, line 150:
```suggestion
    # NOTE: was -2 because for order=0 the first two coefficients disappear as sin(0) == 0
    # removed this now as we no longer fit beta0 and beta1
    n_params = order * 4
```

(I don't like if comments lead to code wraps)

mesmer/mesmer_m/harmonic_model.py
```diff
@@ -152,65 +144,79 @@ def calculate_bic(n_samples, n_order, mse):
...
+        mse = np.mean((predictions - monthly_target) ** 2)
 
-        bic_score[i_n - 1] = calculate_bic(len(y), i_n, mse)
+        bic_score[i_n - 1] = calculate_bic(len(monthly_target), i_n, mse)
+        # TODO: we could stop fitting when we hit a minimum, similar to _minimize_local_discrete
 
-    n_sel = np.argmin(bic_score) + 1
-    coeffs_fit, y_pred = fit_fourier_series_np(x, y, n_sel, repeat=repeat)
+    selected_order = np.argmin(bic_score) + 1
+    # TODO: we already fit for this order above so it would probably be faster to save the output from above and return it?
```
mathause, line 190: :+1: (for later - also my comment re the start values)

mathause (approved):
I added some comments but they are for later (except the code suggestion).

mesmer/mesmer_m/harmonic_model.py
mathause: see #422

PR #426, opened 2024-04-11
fix function call in power transformer

veni-vidi-vici-dormivi:
Fix mistake made in #425.


veni-vidi-vici-dormivi:
This is what happens when you don't test and Vici overrides the review when Mathias is on holiday.

PR #427, opened 2024-04-11
Fix bounds for lambda optimization

veni-vidi-vici-dormivi:
We change the bounds to be consistent with previous MESMER-M code.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@ ci:
     autoupdate_schedule: monthly
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.5.0
+    rev: v4.6.0
     hooks:
       - id: trailing-whitespace
       - id: end-of-file-fixer
@@ -21,7 +21,7 @@ repos:
       - id: isort
   # https://github.com/python/black#version-control-integration
   - repo: https://github.com/psf/black
-    rev: 24.3.0
+    rev: 24.4.2
     hooks:
       - id: black
       - id: black-jupyter
--- a/mesmer/core/utils.py
+++ b/mesmer/core/utils.py
@@ -108,7 +108,7 @@ def _assert_annual_data(time):
         )
 
 
-def upsample_yearly_data(yearly_data, monthly_time):
+def upsample_yearly_data(yearly_data, monthly_time, time_dim="time"):
     """Upsample yearly data to monthly resolution by repeating yearly values.
 
     Parameters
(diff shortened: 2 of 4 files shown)


mathause (commented):
- Does the comment above the changed line need updating (or deletion)?
- I assume that are bounds of `[0, 1]` and `[0, 0.1]`? If so it would be more clear to write as `np.array([[0, 1], [0, 0.1]])` (and easier to understand than `c_[...]`)

veni-vidi-vici-dormivi:
Okay so by now I understand more of this. So let's start over:
The Yeo-Johnson transform aims to transform tailed distributions to more symmetric distributions.
image-20240430143126199
🔼 For a left tailed distribution like the above for example, we need a convex function, that contracts the longer tail and expands the shorter tail (please ignore that the function is too flat towards the y axis, this should be steeper). Lambdas between 1 and 2 make for a convex transformation function like the above (the higher lambda, the more curved), lambdas between 0 and 1 for a concave function.

A lambda of 1 means no change to the distribution, equivalent to the data being normally distributed.

We now want to find optimal lambdas to transform the monthly residuals after the harmonic model to be more normally distributed. Herein, we assume (or Shruti found) that it makes sense that lambda is a function of the local yearly temperature. This confused be at first, because one observation (one local monthly temperature residual of a single year) does not make for a great distribution. However, the underlying assumption is that, if we had more realizations of the local monthly temperature residual of each year, the (skewness of the) distribution of them would depend on the yearly average local temperature. And the way in which it does  is the same for all the years. Thus, we find lambda as a function of the yearly local temperature by fitting for the parameters in the function:
$\lambda_{m,s,y} = \frac{2}{1+ \xi_{0,m,s} \cdot e^{\xi_{1,m,s} \cdot T_{s,y}}}$

Which is a logistic function ranging between 0 and 2 if $\xi_0$ is positive. then, $\xi_0$ controls the intercept, $\xi_1$ the slope. Importantly, when the given data is normally distributed for all yearly temperatures, the parameters should converge towards 1 and 0, giving us a lambda of one for all yearly temperatures. Thus, I see no reason why $\xi_0$ should be bound between 0 and 1, especially because we expect a value that is close to one for normally distributed data. So I am updating the bounds for $\xi_0$ to $[0, \infty]$ (intercept is at 2 for 0 and goes towards 0 for xi towards infinity) because that makes more physical sense to me. As for the bounds on $\xi_1$ I don't really see an obvious reason to put a bound on it at all except for maybe performance when fitting? but I would definitely opt to have negative values in there because having just positive ones would mean that positive yearly temperature anomalies would always lead to more right skewed monthly residuals and vice versa for negative anomalies, which I don't think is true? 

Also I want to update the first guess to [1, 0] in another pull request, so our first guess would be that the data already is normally distributed, which is a more physical guess than the current one (0.01 and 0.01) and also closer to the real results.

mathause (approved):
This change makes sense in any case, so let's do this. 

We can continue the discussion on the power transformer in #440.

veni-vidi-vici-dormivi:
You are right, I accidentally overwrote the changes when I resolved the merge conflict. 🤦‍♀️

PR #429, opened 2024-04-30
factor out and adjust firstguess in lambda fit

veni-vidi-vici-dormivi:
See https://github.com/MESMER-group/mesmer/pull/427#issuecomment-2085316027

PR #430, opened 2024-04-30
Add tests for Power Transformer

veni-vidi-vici-dormivi:


veni-vidi-vici-dormivi:
So these fail because the bounds and first guess are not good. Hopefully, after merging #429 and #427 in here, the tests pass. They do locally with the new bounds and fg.

veni-vidi-vici-dormivi:
Hm why are we not getting a codecov report here?

mathause:
> Hm why are we not getting a codecov report here?

The log gives a hint:https://github.com/MESMER-group/mesmer/pull/430/checks#step:8:55

Can't say that I understand it, though 🤭. I thought codecov was moving away from tokens, but they seem to recommend it again. Let me check what xarray does... Also why only this PR and why consistently?

mathause:
Ah I think it's here: https://docs.codecov.com/docs/codecov-uploader#supporting-token-less-uploads-for-forks-of-open-source-repos-using-codecov (the last paragraph)

tests/unit/test_power_transformer.py
```diff
@@ -0,0 +1,80 @@
...
+def test_lambda_function():
+    # Note that we test with normally distributed data
+    # which should make lambda close to 1 and
+    # the coefficients close to 1 and 0
+    # but for the sake of testing, we set the coefficients differently
+    coeffs = [1, 0.1]
+    local_yearly_T_test_data = np.random.rand(10) * 100
+
+    # even for random numbers, the lambdas should always be between 0 and 2
+    lambdas = lambda_function(coeffs, local_yearly_T_test_data)
```
mathause, line 18: What does `lambda_function` actually do?

tests/unit/test_power_transformer.py
```diff
@@ -0,0 +1,80 @@
...
+def test_fit_power_transformer():
+    # with enough random data points the fit should be close to 1 and 0
+    gridcells = 1
+    years = 100000
+    monthly_residuals = np.random.rand(years, gridcells) * 10
+    yearly_T = np.ones((years, gridcells))
+
+    pt = PowerTransformerVariableLambda()
+    pt.fit(monthly_residuals, yearly_T, gridcells)
+    pt.coeffs_
```
mathause, line 32:
```suggestion
```

tests/unit/test_power_transformer.py
```diff
@@ -0,0 +1,80 @@
...
+
+    result = pt.coeffs_[0]
+    expected = np.array([1, 0])
+
+    np.testing.assert_allclose(result, expected, rtol=1e-10, atol=1e-10)
```
mathause, line 37: Isn't the default tol larger? It's certainly ok to require more precise numbers but I would probably just keep them at the default.

tests/unit/test_power_transformer.py
```diff
@@ -0,0 +1,80 @@
...
+    gridcells = 1
+    years = 100000
+    monthly_residuals = np.random.rand(years, gridcells) * 10
+    yearly_T = np.ones((years, gridcells))
+
+    pt = PowerTransformerVariableLambda()
+    pt.fit(monthly_residuals, yearly_T, gridcells)
+    pt.coeffs_
+
+    result = pt.coeffs_[0]
```
mathause, line 34: How does the full `coeffs_` look?

tests/unit/test_power_transformer.py
```diff
@@ -0,0 +1,80 @@
...
+    np.testing.assert_allclose(result, expected, rtol=1e-10, atol=1e-10)
+
+
+def test_transform():
+    pt = PowerTransformerVariableLambda()
+    pt.standardize = False
+    pt.coeffs_ = np.tile([1, 0], (10, 1))
+
+    monthly_residuals = np.ones((10, 10))
+    yearly_T = np.zeros((10, 10))
```
mathause, line 46: * Factor out `n = 10`?
* Do they have to be symmetric? If not maybe not make them. (Could use `n_ts` and `n_gridpoint` or whatever it is).

tests/unit/test_power_transformer.py
```diff
@@ -0,0 +1,80 @@
...
+    expected = np.ones((10, 10))
+
+    np.testing.assert_equal(result, expected)
+
+
+def test_yeo_johnson_transform():
+    pt = PowerTransformerVariableLambda()
+
+    # test all possible combinations of local_monthly_residuals and lambdas
+    local_monthly_residuals = np.array([0, 1, 0, 1, -1, -1])
```
mathause, line 58: Make these float arrays?

tests/unit/test_power_transformer.py
```diff
@@ -0,0 +1,80 @@
...
+    lambdas = np.array([1, 1, 0, 0, 1, 2])
+
+    result = pt._yeo_johnson_transform(local_monthly_residuals, lambdas)
+    expected = np.array([0, 1, 0, 0, -1, 0])
```
mathause, line 62: I assume `result` is a float array?

tests/unit/test_power_transformer.py
```diff
@@ -0,0 +1,80 @@
...
+
+    np.testing.assert_equal(result, expected)
+
+
+def test_inverse_transform():
```
mathause, line 67: I like the test checking code/ decode invariant holds, but it's a bit pointles with these inputs?
Could also do this the other way round.

mathause (commented):
I had a look from my smartphone...

tests/unit/test_power_transformer.py
veni-vidi-vici-dormivi: https://github.com/MESMER-group/mesmer/blob/03ab48cacc324782f7e12b1a0227abc87b7b95b4/mesmer/mesmer_m/power_transformer.py#L9-L10

Just calculates lambda from the coeffs. Here I test if the lambdas always come out between 0 and 2. Because with a logistic function, they should. This would make 
https://github.com/MESMER-group/mesmer/blob/03ab48cacc324782f7e12b1a0227abc87b7b95b4/mesmer/mesmer_m/power_transformer.py#L229-L231
unnecessary.

In theory, also the parts here for lambda < spacing and lambda-2 < spacing 
https://github.com/MESMER-group/mesmer/blob/03ab48cacc324782f7e12b1a0227abc87b7b95b4/mesmer/mesmer_m/power_transformer.py#L168-L171
would become unnecessary but I would leave it in here because this is the general definition of the yes-johnson transform and if we were to choose another lambda function at some point it would be nice to have that there.

tests/unit/test_power_transformer.py
veni-vidi-vici-dormivi: Yes true, the default `rtol` is `1e-07`, but default `atol = 0`. Fixed that now.

tests/unit/test_power_transformer.py
veni-vidi-vici-dormivi: Has another bracket around the result array, agree that this is confusing, changed the result.

tests/unit/test_power_transformer.py
veni-vidi-vici-dormivi: 🤦‍♀️ 👍

veni-vidi-vici-dormivi:
Ah now also the Codecov report's here :)

veni-vidi-vici-dormivi:
We could add another test, but it is not really code related but rather of scientific nature: transform skewed data and test if it is actually less skewed afterwards. What do you think @mathause ?

tests/unit/test_power_transformer.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def test_fit_power_transformer():
+    # with enough random data points the fit should be close to 1 and 0
+    # here we test with uniform random data because it is quicker to fit
+    # Uniform data is also symmetrically distributed so coefficients
+    # should be close to 1 and 0 as well
+
+    gridcells = 1
+    years = 100000
```
mathause, line 30:
```suggestion
    years = 100_000
```

tests/unit/test_power_transformer.py
mathause: But wikipedia says $\lambda$ can be any real number (and I _think_ the paper says the same) - do you know why it is restricted to `0..2` here?

mathause (commented):
Ok I add some more suggestions after I understand it a bit better. However, this whole thing is a quite a mess (not your PR - the thing overall). I'll need another pass once we cleaned this up some more and we start to see through.

tests/unit/test_power_transformer.py
```diff
@@ -0,0 +1,94 @@
...
+    PowerTransformerVariableLambda,
+    lambda_function,
+)
+
+
+def test_lambda_function():
+    # Note that we test with normally distributed data
+    # which should make lambda close to 1 and
+    # the coefficients close to 1 and 0
+    # but for the sake of testing, we set the coefficients differently
```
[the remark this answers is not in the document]
veni-vidi-vici-dormivi: So I'm not sure if I understand correctly what you mean here. The yearly temperature could have a trend (it does in the scenarios, it doesn't in the historical), but I don't think the coefficients have anything to do with this trend. Do you mean this trend? 
The coefficients kind of introduce a trend into lambda_function but I wouldn't call it trend? We could write. note that we test with normally distributed data without skewness but for the sake of testing we use coefficients wich represent skewness?

tests/unit/test_power_transformer.py
```diff
@@ -0,0 +1,125 @@
+import numpy as np
+import scipy
+
+from mesmer.mesmer_m.power_transformer import (
+    PowerTransformerVariableLambda,
+    lambda_function,
+)
+
+
+def test_lambda_function():
```
mathause, line 10: We don't do any fitting or so here - so there is not really any need to use random data & we can do "deterministic" tests

```python
@pytest.mark.parametrize(
    "coeffs, t, expected",
    [
        ([1, 0.1], -np.inf, 2.0),
        ([1, 0.1], np.inf, 0.0),
        ([1, -0.1], -np.inf, 0.0),
        ([1, -0.1], np.inf, 0.0),
        ([0, 0], 1, 2),
        ([0, 1], 1, 2),
        ([1, 0], 1, 1),
        ([2, 0], 1, 2 / 3),
        ([0, 0], 1, 2),
        ([1, 1], np.log(9), 2 / 10),
    ]
)
def test_lambda_function(coeffs, t, expected):

    result = lambda_function(coeffs, t)
    np.testing.assert_allclose(result, expected)
```

tests/unit/test_power_transformer.py
```diff
@@ -0,0 +1,125 @@
+import numpy as np
+import scipy
```
mathause, line 2:
```suggestion
import scipy as sp
```

tests/unit/test_power_transformer.py
```diff
@@ -0,0 +1,125 @@
...
+
+    pt = PowerTransformerVariableLambda()
+    pt.coeffs_ = pt._yeo_johnson_optimize_lambda(local_monthly_residuals, yearly_T)
+    lmbda = lambda_function(pt.coeffs_, yearly_T)
+    transformed = pt._yeo_johnson_transform(local_monthly_residuals, lmbda)
+
+    assert (lmbda > 1).all() & (lmbda <= 2).all()
+    np.testing.assert_allclose(scipy.stats.skew(transformed), 0, atol=0.01)
+
+    # this fails, need to investigate more
```
mathause, line 61: It does not work because of the upper bound of $\xi_0$ (I mean thinking back on this it makes total sense, because `pt.coeffs_[0]` was hugging the upper bound of 1...). It's un-intuitive but a larger $\xi_0$ yields smaller $\lambda$. If you increase the upper bound it works.

Note - it will work better if you remove `jac=rosen_der` (you get closer to the estimate by `PowerTransformer().fit(local_monthly_residuals.reshape(-1, 1)).lambdas_`). I did only one test for this.

(The `PowerTransformer` uses `sp.optimize.brent` for the fitting with `brack=(-2, 2)` but I think the `brack` just give a starting point for the search in contrast of the `bounds` for `optimize`).

tests/unit/test_power_transformer.py
```diff
@@ -0,0 +1,125 @@
...
+    # pt = PowerTransformerVariableLambda()
+    # pt.coeffs_ = pt._yeo_johnson_optimize_lambda(local_monthly_residuals, yearly_T)
+    # lmbda = lambda_function(pt.coeffs_, yearly_T)
+    # transformed = pt._yeo_johnson_transform(local_monthly_residuals, lmbda)
+
+    # assert (lmbda >= 0).all() & (lmbda <= 1).all()
+    # np.testing.assert_allclose(scipy.stats.skew(transformed), 0, atol=0.01)
+
+
+def test_transform():
```
mathause, line 74:
```suggestion
def test_transform():
    # NOTE: testing trivial transform with lambda = 1
```

tests/unit/test_power_transformer.py
```diff
@@ -0,0 +1,125 @@
...
+    monthly_residuals = np.sin(np.linspace(0, 2 * np.pi, n_ts * n_gridcells)).reshape(
+        n_ts, n_gridcells
+    )
+    yearly_T = np.zeros((n_ts, n_gridcells))
+
+    pt = PowerTransformerVariableLambda()
+    pt.standardize = False
+    # dummy lambdas (since yearly_T is zero lambda comes out to be second coefficient)
+    # we have all cases for lambdas 0 and 2 (special cases), 1 (identity case)
+    # lambda between 1 and 1 and lambda between 1 and 2 for concave and convex cases
```
mathause, line 117:
```suggestion
    # lambda between 0 and 1 and lambda between 1 and 2 for concave and convex cases
```

tests/unit/test_power_transformer.py
```diff
@@ -0,0 +1,125 @@
...
+    pt.coeffs_ = np.array([[0, 0], [0, 1], [0, 2], [0, 0.5], [0, 1.5]])
+    pt.mins_ = np.amin(monthly_residuals, axis=0)
+    pt.maxs_ = np.amax(monthly_residuals, axis=0)
```
mathause, line 120: What do the `mins_` and `maxs_` do?

tests/unit/test_power_transformer.py
```diff
@@ -0,0 +1,125 @@
...
+    local_monthly_residuals = np.array([0.0, 1.0, 0.0, 1.0, -1.0, -1.0])
+    lambdas = np.array([1.0, 1.0, 0.0, 0.0, 1.0, 2.0])
+
+    result = pt._yeo_johnson_transform(local_monthly_residuals, lambdas)
+    expected = np.array([0.0, 1.0, 0.0, np.log1p(1.0), -1.0, -np.log1p(1.0)])
+
+    np.testing.assert_equal(result, expected)
+
+
+def test_inverse_transform():
```
mathause, line 104:
```suggestion
def test_transform_roundtrip():
```

tests/unit/test_power_transformer.py
veni-vidi-vici-dormivi: They are used after the transform to clip the inverted data to the original range. Shruti said this is a relict of when she did not use a logistic lambda function. The she would sometimes get data far outside the original range. As far as I can tell this is not actually needed anymore. However, we should change that somewhere else so here I set the attributes for the PowerTransformer because it fails otherwise.

tests/unit/test_power_transformer.py
mathause: Do you have any objection to this suggestion?

mathause (commented):
You ignored some of my comments, which I think are still valid.

veni-vidi-vici-dormivi:
> You ignored two of my comments, which I think are still valid.

Sorry, hadn't pushed them yet.

veni-vidi-vici-dormivi:
Okay this got messy but now it's ready to merge.

mathause (approved):
Nice! Thanks.

PR #431, opened 2024-04-30
Add tests for harmonic model

veni-vidi-vici-dormivi:


veni-vidi-vici-dormivi:
This fails now because `fit_to_bic_np` still returns nans at the end of the coefficient array but this should be solved after merging #433.

tests/unit/test_harmonic_model.py
```diff
@@ -0,0 +1,115 @@
...
+from mesmer.core.utils import upsample_yearly_data
+from mesmer.mesmer_m.harmonic_model import (
+    fit_to_bic_np,
+    fit_to_bic_xr,
+    generate_fourier_series_np,
+)
+from mesmer.testing import trend_data_2D
+
+
+def test_generate_fourier_series_np():
```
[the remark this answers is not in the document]
veni-vidi-vici-dormivi: No, but I see what you did there. Note that I use `yearly_predictor = np.zeros(n_months)`. That leads to the fact that in your example the factors $3.14$ and $1$ get ignored because they are multiplied with the `yearly_predictor`. But, even if we use `yearly_predictor = np.ones(n_months)` you still forgot to "factor out the brackets?" and actually `expected += 3.14 * np.sin(np.pi * months / 6) + 1 * np.cos(np.pi * months / 6)`. 
I'll add a test using ones and your example.

tests/unit/test_harmonic_model.py
```diff
@@ -0,0 +1,115 @@
...
+    selected_order, estimated_coefficients, predictions = fit_to_bic_np(
+        yearly_predictor, monthly_target, max_order=6
+    )
+
+    # assert selected_order == int(len(coefficients) / 4)
+    # the model does not necessarily select the "correct" order
+    # but the coefficients of higher orders should be close to zero
+
+    # linear combination of the coefficients with the predictor should be similar
+    np.testing.assert_allclose(
```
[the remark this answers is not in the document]
veni-vidi-vici-dormivi: I guess we answered this in https://github.com/MESMER-group/mesmer/pull/433#issuecomment-2112016704.
What could be a nice option to do this, would be to just replace all nans in the estimated coefficients with zeros. This way new can compare them nicely and can still give nans to the user.

veni-vidi-vici-dormivi:
Okay this looks good, now for the last hurdle: how do we want to test numerical stability? I can just hard code some expected output into it or we can save output data somewhere. The thing is that for hard coding output it is rather nasty because it is notoriously hard (impossible?) to get pretty numbers, but for saving it it is rather little data to check... To see what I mean I push an example hard coded test below.

tests/unit/test_harmonic_model.py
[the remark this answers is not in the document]
veni-vidi-vici-dormivi: >  the order would not align with the coeffs

Why? 

> Could you compute the arrays outside of the assert_allclose

Yes definitely, but maybe we won't need it, as of the comment below.

> Isn't this only for the first two coeffs?

hm no?

tests/unit/test_harmonic_model.py
```diff
@@ -0,0 +1,223 @@
...
+
+    yearly_predictor = np.zeros(n_months)
+    months = np.tile(np.arange(1, 13), n_years)
+
+    # dummy yearly cycle
+    expected = -np.sin(2 * np.pi * (months) / 12) - 2 * np.cos(
+        2 * np.pi * (months) / 12
+    )
+    result = generate_fourier_series_np(
+        yearly_predictor, np.array([0, -1, 0, -2]), months
```
mathause, line 28: I just realized it's $a \cdot T + b$ and not $a + b \cdot T$ in the code. That's not confusing at all :man_shrugging:

tests/unit/test_harmonic_model.py
```diff
@@ -0,0 +1,223 @@
...
+    )
+    np.testing.assert_equal(result, expected)
+
+    yearly_predictor = np.ones(n_months)
+    result = generate_fourier_series_np(
+        yearly_predictor, np.array([0, -1, 0, -2]), months
+    )
+    expected += 1
```
mathause, line 36:
```suggestion
    # NOTE: yearly_predictor is added to the Fourier series
    expected += 1
```

tests/unit/test_harmonic_model.py
```diff
@@ -0,0 +1,223 @@
...
+    [
+        np.array([0, -1, 0, -2]),
+        np.array([1, 2, 3, 4, 5, 6, 7, 8]),
+        np.array([1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]),
+    ],
+)
+@pytest.mark.parametrize(
+    "yearly_predictor",
+    [np.repeat([-1, 1], 10 * 6), np.linspace(-1, 1, 10 * 12) * 10],
+)
```
mathause, line 57: Could move that into the function if it's only one (but also ok to leave as is).

tests/unit/test_harmonic_model.py
```diff
@@ -0,0 +1,223 @@
...
+    #         estimated_coefficients[i] * yearly_predictor[i]
+    #         + estimated_coefficients[i + 1]
+    #         for i in range(0, len(estimated_coefficients), 2)
+    #     ],
+    #     atol=1e-2,
+    # )
+
+    np.testing.assert_allclose(original_coefficients, estimated_coefficients, atol=1e-2)
+
+    # but actually all what really counts is that the predictions are close to the target
```
mathause, line 98:
```suggestion
```

mathause (approved):

tests/unit/test_harmonic_model.py
veni-vidi-vici-dormivi: ...but its two?

mathause: Ah yes, sorry

tests/unit/test_harmonic_model.py
mathause: No, I was just quite confused (but true, I have seen both conventions)

PR #433, opened 2024-05-02
allow nans in generate harmonic model sum

veni-vidi-vici-dormivi:
In the harmonic model we return the fit coefficient array with nans, for coefficients, higher than the order we chose. We need to fill the array up until the max order, so we have coefficient arrays of the same length for all gridcells, so filling the array up with nans for unfit coefficients might have made sense in the beginning. However, imo it makes more sense to fill it up with zeros because in this way one can still use the fit coefficient array to `generate_fourier_series` and it makes testing easier, see #431.

### DIFF
--- a/mesmer/mesmer_m/harmonic_model.py
+++ b/mesmer/mesmer_m/harmonic_model.py
@@ -41,7 +41,7 @@ def generate_fourier_series_np(coeffs, order, yearly_T, months):
     beta0 = 0
     beta1 = 1
 
-    seasonal_cycle = sum(
+    seasonal_cycle = np.nansum(
         [
             (coeffs[idx * 4] * yearly_T + coeffs[idx * 4 + 1])
             * np.sin(np.pi * i * (months) / 6)
@@ -193,7 +193,7 @@ def fit_to_bic_np(yearly_predictor, monthly_target, max_order):
     )
 
     # need the coeff array to be the same size for all orders
-    coeffs = np.zeros([max_order * 4])
+    coeffs = np.zeros([max_order * 4]) * np.nan
     coeffs[: selected_order * 4] = coeffs_fit
 
     return selected_order, coeffs, predictions
--- a/mesmer/mesmer_m/harmonic_model.py
+++ b/mesmer/mesmer_m/harmonic_model.py
@@ -193,7 +193,7 @@ def fit_to_bic_np(yearly_predictor, monthly_target, max_order):
     )
 
     # need the coeff array to be the same size for all orders
-    coeffs = np.zeros([max_order * 4]) * np.nan
+    coeffs = np.zeros([max_order * 4])
     coeffs[: selected_order * 4] = coeffs_fit
 
     return selected_order, coeffs, predictions


mathause:
Unsure about this. Before we merge - would it work with `nan` if we update

https://github.com/MESMER-group/mesmer/blob/2fdb3d585f9eaf949f15ef3d325347a0ca1d8827/mesmer/mesmer_m/harmonic_model.py#L45

to `np.nansum(..., axis=)`?

mathause:
(I like the beauty of `0` but `nan` is more explicit.)

veni-vidi-vici-dormivi:
No, because we can'T multiply a sequence with nan.

mathause:
Doesn't that just result in `nan`? Where in the code) is that?

(Not trying to be obnoxious, maybe just a little slow. Ah and nothing weird looking at code on Saturday morning at 7 during my holidays 🙃 (I am waiting for my flight))

mathause (changes requested):
I am quite sure we need `np.nansum(..., axis=0)` (also we will eventually have to think about inefficiencies but can do that later)

veni-vidi-vici-dormivi:
> I am quite sure we need np.nansum(..., axis=0)

Why? We only have data in one axis, it does not matter if we sum over axis 0 or all of them. And since it is only used internally this is guaranteed?

mathause (approved):
Yes, you are right - I forgot about the for loop. Do we need to update the title of the PR?

veni-vidi-vici-dormivi:
So you were right, we need axis = 0 because the list actually gets a new dimension for each order. I didn't reload the notebook properly when testing it. Fixing this and finishing work on the tests next 🤦‍♀️

mathause:
Interesting - it might be necessary to discuss that in person. I hope to be "fit" again soon... 

Just as a small comment - if we don't fit for $a$ and $b$ (but manually set it to $0$ and $1$) we should subtract $T_y$ and fit on the residuals. Would that help here?

veni-vidi-vici-dormivi:
Yeah I think that's a good idea. No, I'm sorry $a$ and $b$ are not the ones that control the power of the trend (we did set them to 0 and 1 manually and don't fit them) but they stand for the coefficients in each amplitude of sine and cosine terms in the Fourier series. And I think we do fit on the residuals (see discussion in #443).

$$
T_{m,s,y} = \beta_{0,s} + \beta_{1,s} \cdot T_{s,y} + \sum_{i = 1}^{n=6} [(a_{1,i,s} + b_{1,i,s} \cdot T_{s,y}) \cdot sin(\frac{i \pi m}{6}) + (a_{2,i,s} + b_{2,i,s} \cdot T_{s,y}) \cdot cos(\frac{i \pi m}{6})] 
$$

mathause:
Oh right I think I understand now. So given $c = (a + b \cdot T)$ you say that there are different combinations of $a_1$ and $b_1$ leading to the same $c$?

veni-vidi-vici-dormivi:
Exactly.

Issue #410, opened 2024-03-04
optimize `_generate_fourier_series_np`


mathause:
I think the mesmer_m part can be optimized. For example `generate_fourier_series_np`.

1. I think the name is not optimal. That should probably be `estimate` or `predict` instead of `generate`.
2. The code can be sped up by vectorizing it:
   ```python
       # TODO:
       # - can remove n from the inputs as it's coeffs.shape / 4
       # - rename n to order (as n is not a good name, but see above)
       # - why is there a "-2" for the number of coeffs?
       # - can this be rewritten as fft? (no -> as the amplitude varies with x)
   
       coeffs_ = coeffs.reshape(-1, 4)
   
       k = np.arange(coeffs_.shape[0])[:, np.newaxis]
   
       alpha = 2 * np.pi * k * (mon % 12 + 1) / 12
   
       predictions = np.sum(
           (coeffs_[:, 0:1] * x + coeffs_[:, 1:2]) * np.sin(alpha)
           + (coeffs_[:, 2:3] * x + coeffs_[:, 3:4]) * np.cos(alpha),
           axis=0,
       )
   
       return predictions
   ```

---

And similar for other functions.

veni-vidi-vici-dormivi:
After playing with it for a while I actually think the name is accurate. The real data is not a Fourier series so estimate and predict is not what we really do. We generate a Fourier series for the estimated or predicted coefficients which are estimated/predicted in `fit_fourier_series_np`.

mathause:
I thought about if it makes sense to save the `coeffs` as 2D array, i.e. with dims `order x coeffs` and size `order x 4`. This could help with interpretation.

PR #434, opened 2024-05-02
Infer harmonic model order from coefficients

veni-vidi-vici-dormivi:
In the harmonic model `generate_fourier_series` we passed the order as an argument instead of inferring it from the length of the coefficients array. This is changed here. Additionally, I reordered the arguments to facilitate testing on xarrays.

### DIFF
--- a/mesmer/mesmer_m/harmonic_model.py
+++ b/mesmer/mesmer_m/harmonic_model.py
@@ -14,15 +14,13 @@ from scipy import optimize
 import mesmer
 
 
-def generate_fourier_series_np(coeffs, order, yearly_T, months):
+def generate_fourier_series_np(coeffs, yearly_T, months):
     """construct the Fourier Series
 
     Parameters
     ----------
     coeffs : array-like of shape (4*order)
         coefficients of Fourier Series.
-    order : Integer
-        Order of the Fourier Series.
     yearly_T : array-like of shape (n_years*12,)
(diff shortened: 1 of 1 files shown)


mathause (approved):
I think it's a good idea to rename the function and make it private but that can be another PR (#410).

mesmer/mesmer_m/harmonic_model.py
```diff
@@ -34,7 +32,7 @@ def generate_fourier_series_np(coeffs, order, yearly_T, months):
         Fourier Series of order n calculated over yearly_T and months with coeffs.
 
     """
-    # TODO: infer order from coeffs
+    order = int(len(coeffs) / 4)
```
mathause, line 35:
```suggestion
    order = int(coeffs.size / 4)
```
if it's guaranteed to be a 1D array. (Could also do `// 4` but `int` is more explicit.)

veni-vidi-vici-dormivi: Hm when we make the function private it should be guaranteed that it's 1D. Or do you prefer to implement a check?

mathause: I quickly checked and `len(arr) == arr.shape[0]` - so the assumption is there already, and I also suspect it would fail below. Thus, I suggest we go ahead and commit this suggestion.

PR #435, opened 2024-05-02
fix passing time dimension instead of dataarray in upsampling

veni-vidi-vici-dormivi:


mathause (approved):
Maybe make it explicit?
```python
time_dim = "time"
...
```
but also fine to merge as is.

veni-vidi-vici-dormivi:
You mean implementing a keyword argument in `upsample_yearly_data` and `fit_to_bic_xr` and passing the monthly_data data array (potentially including tas) into upsampling?
I agree that a flexible time dimension would be better.

mathause:
I thought to just write it out for now an let the other be a long term goal

veni-vidi-vici-dormivi:
> let the other be a long term goal
Whoopsie too late. What do you think about this? Then I would change the PR title to "Make time dim explicit in upsampling yearly data".

PR #436, opened 2024-05-02
align equality check in yeo johnson transform

veni-vidi-vici-dormivi:
Changing the comparison to eps in `_yeo_johnson_transform` to be consistent with [sklearn's yet-johnsons power transform](https://github.com/scikit-learn/scikit-learn/blob/8721245511de2f225ff5f9aa5f5fadce663cd4a3/sklearn/preprocessing/_data.py#L3396).

### DIFF
--- a/mesmer/mesmer_m/power_transformer.py
+++ b/mesmer/mesmer_m/power_transformer.py
@@ -165,7 +165,7 @@ class PowerTransformerVariableLambda(PowerTransformer):
 
         transformed = np.zeros_like(local_monthly_residuals)
         # get positions of four cases:
-        # NOTE: this code is copied from sklearn's PowerTransformer, see 
+        # NOTE: this code is copied from sklearn's PowerTransformer, see
         # https://github.com/scikit-learn/scikit-learn/blob/8721245511de2f225ff5f9aa5f5fadce663cd4a3/sklearn/preprocessing/_data.py#L3396
         # we acknowledge there is an inconsistency in the comarison of lambdas
         pos_a = (local_monthly_residuals >= 0) & (np.abs(lambdas) < eps)
--- a/mesmer/mesmer_m/power_transformer.py
+++ b/mesmer/mesmer_m/power_transformer.py
@@ -165,6 +165,9 @@ class PowerTransformerVariableLambda(PowerTransformer):
 
         transformed = np.zeros_like(local_monthly_residuals)
         # get positions of four cases:
+        # NOTE: this code is copied from sklearn's PowerTransformer, see 
+        # https://github.com/scikit-learn/scikit-learn/blob/8721245511de2f225ff5f9aa5f5fadce663cd4a3/sklearn/preprocessing/_data.py#L3396
+        # we acknowledge there is an inconsistency in the comarison of lambdas
         pos_a = (local_monthly_residuals >= 0) & (np.abs(lambdas) < eps)
         pos_b = (local_monthly_residuals >= 0) & (np.abs(lambdas) >= eps)
         pos_c = (local_monthly_residuals < 0) & (np.abs(lambdas - 2) > eps)
--- a/mesmer/mesmer_m/power_transformer.py
+++ b/mesmer/mesmer_m/power_transformer.py
@@ -161,8 +161,8 @@ class PowerTransformerVariableLambda(PowerTransformer):
 
         transformed = np.zeros_like(local_monthly_residuals)
         # get positions of four cases:
-        pos_a = (local_monthly_residuals >= 0) & (np.abs(lambdas) <= np.spacing(1.0))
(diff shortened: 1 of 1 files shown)


mathause:
Can you define

```python
eps = np.spacing(np.float64(1))
```
or actually, if we don't do it for an array
```python
eps = np.finfo(np.float64).eps
```
?
That would be clearer to me (not sure if here or in another PR, though).

veni-vidi-vici-dormivi:
I think another PR would make sense.

mathause:
Can you merge main - I need to think about this.

veni-vidi-vici-dormivi:
The question is, if when lambda is exactly eps, if we consider that to be zero, or not. I just wanted to make it consistent with the sklearn function. They however are not consistent themselves I think because they write `if abs(lmbda) < np.spacing(1.0)` for $\lambda = 0$ but then `if not abs(lmbda - 2) > np.spacing(1.0)` for $\lambda = 2$...

```
# when x >= 0
if abs(lmbda) < np.spacing(1.0):
    out[pos] = np.log1p(x[pos])
else:  # lmbda != 0
    out[pos] = (np.power(x[pos] + 1, lmbda) - 1) / lmbda

# when x < 0
if abs(lmbda - 2) > np.spacing(1.0):
    out[~pos] = -(np.power(-x[~pos] + 1, 2 - lmbda) - 1) / (2 - lmbda)
else:  # lmbda == 2
    out[~pos] = -np.log1p(-x[~pos])
```

Maybe that's the reason why Shruti wrote it herself? Because for the inverse transform she actually just uses the once from sklearn...

https://github.com/MESMER-group/mesmer/blob/03ab48cacc324782f7e12b1a0227abc87b7b95b4/mesmer/mesmer_m/power_transformer.py#L272

mathause:
I Shruti wrote this herself so she could have variable (or dependent) $\lambda$ values.

I think the `>=` / `>` mess comes because the original author also confused that $\lambda$ can be any real value and does not have to be in `0..2` (also https://github.com/MESMER-group/mesmer/pull/430#discussion_r1590938808). The value problem was fixed but this particular inconsistency remained - see scikit-learn/scikit-learn#12522 (_Assuming_ `0 <= lambda <= 2` the operators make sense.)

veni-vidi-vici-dormivi:
> Shruti wrote this herself so she could have variable (or dependent)  values. 

Hm, I mean we could also pass every (value, lambda) pair to the sklearn power transform no? Like we do for the inverse transform.

> $\lambda$ can be any real value and does not have to be in 0..2
In our case it is because lambda is derived from a logistic function.

veni-vidi-vici-dormivi:
I don't get how it makes sense that in one case $\lambda == eps$ is not contained in the case and in the other it is?

mathause:
> > Shruti wrote this herself so she could have variable (or dependent)  values.
> 
> Hm, I mean we could also pass every (value, lambda) pair to the sklearn power transform no? Like we do for the inverse transform.

Yes, but then we have to check if this is vectorized - otherwise it will be too slow.

> 
> > λ can be any real value and does not have to be in 0..2
> 
> In our case it is because lambda is derived from a logistic function.

Ah ok, sorry - it's too many open PRs and comments. But then this is a function of the covariate function and it's not optimal if this is in

https://github.com/MESMER-group/mesmer/blob/10c8c432ba21b98d75dd05fbfe70b7d7d30404d5/mesmer/mesmer_m/power_transformer.py#L219

(Technically the user will not be able to easily replace the `lambda_function` but it's misleading when the bounds are associated with the Yeo-Johnson transform and not with the covariate function. And almost impossible to find out if the `lambda_function` should ever be changed...)

veni-vidi-vici-dormivi:
I agree, it is pretty hard to see through it all. The `get_yeo_johnson_lambdas` is pretty horrible. I think we could get rid of it when rewriting the whole thing for xarrays. Maybe we should think about passing the lambda function as argument... 
But in this PR I actually just wanted to fix the operator, to be the same as in sklearn. Am I missing something?

mathause:
> But in this PR I actually just wanted to fix the operator, to be the same as in sklearn. Am I missing something?

No - I was trying to understand why it's inconsistent in scikit-learn (and I think I do now).

mathause (approved):
Ok, some more comments.

- I don't think what we choose will matter _in practice_.
- `eps` is much smaller for 0 than for 1
   ```python
   np.spacing([0, 1, 2])
   ```
   ```
   array([4.94065646e-324, 2.22044605e-016, 4.44089210e-016])
   ```
- That means there are values between 0 and eps (otherwise `np.abs(lambdas) < eps` would be equal to `np.abs(lambdas) == 0`)

---

Ok, let's merge this but I think we need to fix the title (we don't change a sign). Maybe: "align equality check in yeo johnson transform"?

Issue #348, opened 2023-12-05
mesmer module naming and structure


mathause:
I feel like I am being held up by this bike-shedding problem and I would like to settle this before releasing the next version. @znicholls, @jschwaab & @yquilcaille - do you have an opinion?

### 1. module name for climate data processing?

The statistical functions are collected under `mesmer.stats.*`. I like this because its short and descriptive. So far I have failed to come up with a similarly good name for the climate data processing. Currently these functions are under

```python
mesmer.geospatial.*
mesmer.grid.*
mesmer.mask.*
mesmer.weighted.*
```

It feels a bit asymmetric... **Should these be moved to a common module? If yes - how should it be called?** I list some ideas, although none entirely convinces me (**happy to take suggestions**)...

```python
mesmer.cdp.*       # climate data processing -> makes sense, does not read so well
mesmer.cdt.*       # climate data tools      -> makes sense, does not read so well
mesmer.xcm.*       # xr climate model        -> makes only a bit sense, reads well
mesmer.xcmd.*      # xr climate model data   -> makes a bit more sense, reads a bit less well
# while these are more expressive they seem rather long 
mesmer.dataproc.*
mesmer.data_processing.*
mesmer.climate_data_tools.*
mesmer.clim_dta_tools.*
mesmer.process.*
```

Similar existing libraries & programs are cdo, xESMF, xgcm, xmip... 

### 2. where should public functions be exposed?

Currently functions etc. are exposed in their own module, e.g. `mesmer.stats.auto_regression.fit_auto_regression`. This has the advantage that there is a better separation between them. However, for many functions, there is duplication in the name (e.g. `2 x "auto_regression"`) and we need good names for all of them. Alternatively we could expose them as `mesmer.stats.fit_auto_regression`. Thus, the question - **should all functions be moved one level down?**

Instead

```python
mesmer.stats.linear_regression.LinearRegression()
mesmer.stats.smoothing.lowess(...)
mesmer.stats.auto_regression.select_ar_order(...)
mesmer.stats.auto_regression.fit_auto_regression(...)
```

we would go with

```python
mesmer.stats.LinearRegression()
mesmer.stats.lowess(...)
mesmer.stats.select_ar_order(...)
mesmer.stats.fit_auto_regression(...)
```

yquilcaille:
I agree that the module stats would be much bigger than each of these four ones. In those that you suggest, xcmd is not bad. Following our discussion, i suggest as well:
[...]

PR #442, opened 2024-05-07
Write Power Transformer functions for xarray compatibility

veni-vidi-vici-dormivi:
Here I write functions for the power transformer to take in xarrays.

 - [ ] Tests added


veni-vidi-vici-dormivi:
At the moment I am appending functions for the xarray power transformer below the old functions. I also now opted to write functions instead of class methods. This is WIP though we can build on this first draft here.

mesmer/mesmer_m/power_transformer.py
```diff
@@ -290,3 +291,233 @@ def inverse_transform(self, transformed_monthly_T, yearly_T):
...
+            _yeo_johnson_optimize_lambda,
+            monthly_residuals_grouped[month],
+            yearly_T,
+            input_core_dims=[["time"], ["time"]],
+            output_core_dims=[[], []],
+            output_dtypes=[float, float],
+            vectorize=True,
+            join="outer",
+        )
+        data = {"xi_0": xi_0, "xi_1": xi_1}
```
mathause, line 449:
```suggestion
        data_vars = {"xi_0": xi_0, "xi_1": xi_1}
```

(or create the `Dataset` directly)

mesmer/mesmer_m/power_transformer.py
```diff
@@ -290,3 +291,233 @@ def inverse_transform(self, transformed_monthly_T, yearly_T):
...
+
+def _get_lambdas_from_covariates_xr(coeffs, yearly_T):
+
+    lambdas = 2 / (1 + coeffs.xi_0 * np.exp(yearly_T * coeffs.xi_1))
+
+    return lambdas.rename("lambdas")
+
+
+def fit_power_transformer_xr(monthly_residuals, yearly_T):
+    """Estimate the optimal parameter lambda for each gridcell, given an xarray dataset
```
mathause, line 416: Careful using `dataset` as a generic name

mathause (commented):
- Can you fix the docstrings and comments so they no wider than the 88 characters.
- We need to align the names of the main functions - not sure if I prefer `power_transformer` or `yeo_johnson`... On second thought I think `yeo_johnson` because `power_transformer` includes the box-cox transformation. Currently they are not consistent:
  ```python
  power_transformer.fit_power_transformer_xr(...)
  power_transformer.yeo_johnson_transform_xr(...)
  power_transformer.inverse_transform(...)
  ```
- Can still consider what gives the better naming & usability - 3 functions or a class (no worries this would not mean much additional work, as the functions are `@staticmethod`)
- What is your suggestion for the next steps i.e. getting rid of the `PowerTransformerVariableLambda` class? (I thought about advising to rip out the internals and replace them with your new functions, but I am less sure if we remove the whole class next)

veni-vidi-vici-dormivi:
> Can you fix the docstrings and comments so they no wider than the 88 characters.

Does the precommit hook not work for docstrings and comments?

mathause:
Yes I am no longer happy about the approach I took for the linear regression - especially the  `lr.params = params` part. Is creating a class just a cheap way around the flatter hierarchy we discussed in #348 (i.e. having everything in the `mesmer.stats` namespace)? But then again - having 3-4 functions for each method makes it cluttered fast... Really not sure what is best... we may have to re-visit #348 at some point...

Sticking with functions for now, maybe?

```python
mesmer.stats.yeo_johnson_fit(...)
mesmer.stats.yeo_johnson_transform(...)
mesmer.stats.yeo_johnson_inverse_transform(...)
```

tests/unit/test_power_transformer.py
```diff
@@ -151,3 +154,92 @@ def test_standard_scaler():
...
+    pt_coefficients = power_transformer.fit_yeo_johnson_transform(
+        monthly_residuals, yearly_T
+    )
+    transformed = power_transformer.yeo_johnson_transform_xr(
+        monthly_residuals, pt_coefficients, yearly_T
+    )
+    inverse_transformed = power_transformer.inverse_yeo_johnson_transform_xr(
+        transformed.transformed, pt_coefficients, yearly_T
+    )
+    xr.testing.assert_allclose(inverse_transformed.inverted, monthly_residuals)
```
mathause, line 205: Add checks to ensure the shape and size is correct?

mathause (approved):
Thanks - I think that's basically ready.

One thing we need to think about at one point - how do we best store coeffs in a Dataset/ DataArray - is it better to have individual arrays (as here or for lin reg) or an array of params? (It probably depends on the kind of coeffs...)

---


Can you run velin (a docstring checker) on the file?

```bash
pip install velin

velin mesmer/mesmer_m/power_transformer.py
# velin mesmer/mesmer_m/power_transformer.py --write
```

(Can also run it on the full project, but prob. in another PR, also `velin .` will throw an error because it picks up the .git folder, so need to do `velin mesmer`.)

PR #443, opened 2024-05-07
Harmonic model only fit orders until local minimum

### LINKED ISSUE #422: MESMER-M: use coeffs from `order - 1` as first guess
To fit the harmonic model, MESMER-M currently uses 0 as first guess. We could start from the coefficients from the previous order instead.

```python
first_guess = np.zeros(order * 4)
first_guess[:(order - 1) * 4] = coeffs
```

veni-vidi-vici-dormivi:
- [ ] Closes #422 

This brings down the time needed to fit the harmonic model from 40.8 seconds to 8.9 seconds.

### DIFF
--- a/mesmer/mesmer_m/harmonic_model.py
+++ b/mesmer/mesmer_m/harmonic_model.py
@@ -51,7 +51,7 @@ def generate_fourier_series_np(yearly_T, coeffs, months):
     return beta0 + beta1 * yearly_T + seasonal_cycle
 
 
-def fit_fourier_series_np(yearly_predictor, monthly_target, order):
+def fit_fourier_series_np(yearly_predictor, monthly_target, order, first_guess):
     """execute fitting of the harmonic model/fourier series
 
     Parameters
@@ -65,6 +65,9 @@ def fit_fourier_series_np(yearly_predictor, monthly_target, order):
 
     order : Integer
         Order of the Fourier Series.
+
+    first_guess : array-like of shape (4*order)
+        Initial guess for the coefficients of the Fourier Series.
 
     Method
     ------
@@ -89,7 +92,7 @@ def fit_fourier_series_np(yearly_predictor, monthly_target, order):
     # get monthly values
     mon_train = np.tile(np.arange(1, 13), int(yearly_predictor.shape[0] / 12))
 
-    def func(coeffs, order, yearly_predictor, mon_train, mon_target):
+    def func(coeffs, yearly_predictor, mon_train, mon_target):
         """loss function for fitting fourier series in scipy.optimize.least_squares"""
 
         loss = np.mean(
@@ -102,7 +105,6 @@ def fit_fourier_series_np(yearly_predictor, monthly_target, order):
 
         return loss
 
-    first_guess = np.zeros(order * 4)
 
     # NOTE: this seems to select less 'orders' than the scipy one
     # np.linalg.lstsq(A, y)[0]
@@ -110,7 +112,7 @@ def fit_fourier_series_np(yearly_predictor, monthly_target, order):
     coeffs = optimize.least_squares(
         func,
         first_guess,
-        args=(order, yearly_predictor, mon_train, monthly_target),
+        args=(yearly_predictor, mon_train, monthly_target),
         loss="cauchy",
     ).x
 
@@ -169,30 +171,29 @@ def fit_to_bic_np(yearly_predictor, monthly_target, max_order):
 
     """
 
-    bic_score = np.zeros([max_order])
+    current_min_score = float("inf")
+    last_coeffs = []
+    selected_order = 0
 
     for i_order in range(1, max_order + 1):
 
-        _, predictions = fit_fourier_series_np(
-            yearly_predictor, monthly_target, i_order
+        coeffs, predictions = fit_fourier_series_np(
+            yearly_predictor, monthly_target, i_order, first_guess=np.append(last_coeffs, np.zeros(4))
         )
-        # TODO: in fit_fourier_series_np we already calculate mse, we could just return it and not do it again here?
-        mse = np.mean((predictions - monthly_target) ** 2)
+        
+        mse = np.mean((monthly_target - predictions) ** 2)
+        bic_score = calculate_bic(len(monthly_target), i_order, mse)
 
-        bic_score[i_order - 1] = calculate_bic(len(monthly_target), i_order, mse)
-        # TODO: we could stop fitting when we hit a minimum, similar to _minimize_local_discrete
-
-    selected_order = np.argmin(bic_score) + 1
-    # TODO: we already fit for this order above so it would probably be faster to save the output from above and return it?
-    coeffs_fit, predictions = fit_fourier_series_np(
-        yearly_predictor=yearly_predictor,
-        monthly_target=monthly_target,
-        order=selected_order,
-    )
+        if bic_score < current_min_score:
+            current_min_score = bic_score
+            last_coeffs = coeffs
+            selected_order = i_order
+        else:
+            break
 
     # need the coeff array to be the same size for all orders
     coeffs = np.zeros([max_order * 4]) * np.nan
-    coeffs[: selected_order * 4] = coeffs_fit
+    coeffs[: selected_order * 4] = last_coeffs
 
     return selected_order, coeffs, predictions
 


veni-vidi-vici-dormivi:
We could also write a function for finding the local minimum similar to `minimize_local_discrete`. I wondered if we could use this one here but I would like to preserve all the function output of `fit_fourier_series_np` and `minimize_local_discrete` is not made for that, and we also don't need it in its original use case. 

What do you think @mathause, leave it like this, factor it out to a separate minimize function, rewrite the original `minimize_local_discrete` function or use the origin `minimize_local_discrete` function and get the coeffs and predictions in another call?

mesmer/mesmer_m/harmonic_model.py
```diff
@@ -102,23 +105,24 @@ def func(coeffs, order, yearly_predictor, mon_train, mon_target):
...
-    first_guess = np.zeros(order * 4)
-
     # NOTE: this seems to select less 'orders' than the scipy one
     # np.linalg.lstsq(A, y)[0]
 
-    coeffs = optimize.least_squares(
+    minimize_result = optimize.least_squares(
         func,
         first_guess,
         args=(order, yearly_predictor, mon_train, monthly_target),
```
mathause, line 114:
```suggestion
        args=(yearly_predictor, mon_train, monthly_target),
```

mesmer/mesmer_m/harmonic_model.py
```diff
@@ -102,23 +105,24 @@ def func(coeffs, order, yearly_predictor, mon_train, mon_target):
...
         loss="cauchy",
```
mathause, line 115:
```suggestion
        # TODO: this has currently no influence as `func` returns the loss
        # and not the residuals
        # loss="cauchy",
```

mesmer/mesmer_m/harmonic_model.py
```diff
@@ -51,7 +51,7 @@ def generate_fourier_series_np(yearly_T, coeffs, months):
     return beta0 + beta1 * yearly_T + seasonal_cycle
 
 
-def fit_fourier_series_np(yearly_predictor, monthly_target, order):
+def fit_fourier_series_np(yearly_predictor, monthly_target, order, first_guess):
```
mathause, line 54: Order is no longer needed now, should prob add a check `

```python
if not order * 4 == first_guess.size:
    raise ValueError("")
```

(as this is in the outer loop I think is fine)

mesmer/mesmer_m/harmonic_model.py
```diff
@@ -102,23 +105,24 @@ def func(coeffs, order, yearly_predictor, mon_train, mon_target):
 
         return loss
 
-    first_guess = np.zeros(order * 4)
-
     # NOTE: this seems to select less 'orders' than the scipy one
     # np.linalg.lstsq(A, y)[0]
 
-    coeffs = optimize.least_squares(
+    minimize_result = optimize.least_squares(
```
mathause, line 111:
```suggestion
    minimize_result = sp.optimize.least_squares(
```

mathause (commented):
* I first thought to allow `first_guess=None` but as (if) this function is internal I think it is fine.
* the way it is currently set up (returning the loss and not the residuals), `loss="cauchy"` has no effect - can you confirm? (also cauchy does not seem to be recommended)

mathause:
> We could also write a function for finding the local minimum similar to `minimize_local_discrete`. I wondered if we could use this one here but I would like to preserve all the function output of `fit_fourier_series_np` and `minimize_local_discrete` is not made for that, and we also don't need it in its original use case.
> What do you think @mathause, leave it like this, factor it out to a separate minimize function, rewrite the original `minimize_local_discrete` function or use the origin `minimize_local_discrete` function and get the coeffs and predictions in another call?

- Careful - you can only abort the loop early if it is strictly monotonic - can you confirm this?
- Yes, would be nice if we could use `minimize_local_discrete` but I don't see an easy way, so lets stay here.

mathause:
(I should maybe not have reviewed this PR, but thought it would be an easy win...)

veni-vidi-vici-dormivi:
> the way it is currently set up (returning the loss and not the residuals), loss="cauchy" has no effect - can you confirm? (also cauchy does not seem to be recommended)

Actually, no, it does have an influence... could it be that this is actually wrong? The documentation says func should return the residuals. Could it be that in this way the loss is interpreted as a residual?

mathause:
I agree and think returning the residuals is the correct thing..., however, this sentence in the docstring had me assume returning the loss is fine as well. https://github.com/scipy/scipy/blob/7dcd8c59933524986923cde8e9126f5fc2e6b30b/scipy/optimize/_lsq/least_squares.py#L265

> ```
> It must allocate and return a 1-D array_like of shape (m,) or a scalar.
> ```
 

---

I assumed `loss="cauchy"` would have no influence as `np.log1p(z)` should be monotonic for any scalar z (is that the right word?). I.e. $x > y \implies \ln{}(x) > \ln{}(y)$. However, the first and second derivative also flows into that (`rho[1]` and `rho[2]`) - maybe that has an influence. Or is it the $+1$ part of `np.log1p(z)`?

https://github.com/scipy/scipy/blob/7dcd8c59933524986923cde8e9126f5fc2e6b30b/scipy/optimize/_lsq/least_squares.py#L190

mesmer/mesmer_m/harmonic_model.py
veni-vidi-vici-dormivi: Hm I think we would make `fit_fourier_series_np` private and then we do not expose the first guess to the user anywhere. So I guess a check is not needed? Also we would need additional arguments then either in `fit_fourier_series_np` (order) or in `fit_to_bic_np`(first_guess). Do I understand that right?

mesmer/mesmer_m/harmonic_model.py
mathause:
```suggestion
        # TODO: double check loss function (cauchy seems not recommended,
        # see least_squares docstring)
        loss="cauchy",
```

mesmer/mesmer_m/harmonic_model.py
veni-vidi-vici-dormivi: I leave this for another PR where we go over all the functions in mesmer-m and make them private or not, and edit the init file (and possibly move them into the stats folder, but that can be yet another PR).

mesmer/mesmer_m/harmonic_model.py
```diff
@@ -102,23 +105,24 @@ def func(coeffs, order, yearly_predictor, mon_train, mon_target):
...
         func,
```
mathause, line 112:
```suggestion
        # TODO: func should return the residuals
        func,
```

mathause (approved):
Ok let's add the two `TODO` comments and merge.

veni-vidi-vici-dormivi:
> I assumed loss="cauchy" would have no influence as np.log1p(z) should be monotonic for any scalar z (is that the right word?). I.e. . However, the first and second derivative also flows into that (rho[1] and rho[2]) - maybe that has an influence. Or is it the  part of np.log1p(z)?

Hm the first derivative is not monotonic so maybe. I am not exactly sure why it comes to different results. I made a few observations:
- returning the loss and using cauchy or not leads to different coefficients at an order of $10^{-5}$ and different predictions at $10{-4}$ and the returned MSE is the same for both cases (at least up to $10{-8}$).
- returning the residuals without using cauchy (using just the default loss function so linear), the coefficients and predictions are similar to the ones when returning the loss ($O(10^{-4})$ and $O(10^{-3})$) also with the same MSE as the other two (1.012).
- returning the residuals and using cauchy yields quite different coefficients (differing to the others at $10^{-3}$ and predictions at $10^{-1}$ and a greater MSE than all the others (1.016).
- returning the residuals and using Huber instead of cauchy gets us an MSE of 1.014.

What it does in the background when we return residuals is apply the loss function to the residuals before computing MSE, and cauchy dampens outliers. I'm not sure though what it does when `func` returns a scalar. Thus I would say returning the residuals and using linear loss func is the best option. 

However I found it nice to have the MSE of the chosen parameters returned by the optimize result, which we lose when we do it like that. `OptimizeResult` contains the final value of the cost function which is however MSE(lossfunc(resids)) and not MSE(resids). Computationally though it's not so bad I guess calculating the MSE twice. However, in that case I would find it prettier to pass the residuals to `calculate_bic`and calculate MSE inside of it instead of calculating the MSE in `fit_to_bic` between fitting the Fourier series and calculating the BIC.

> Ok let's add the two TODO comments and merge.

I'll add the TODOs (including some from the stuff above) but I want to wait with merging after we merged the tests.

PR #444, opened 2024-05-13
rename i and j in power transformer transform & inverse transform

veni-vidi-vici-dormivi:
Renamed i and j indexes to gridcell and year for better readability in `PowerTransformerVariableLambda.transform()` and `inverse_transform`.

### DIFF
--- a/mesmer/mesmer_m/power_transformer.py
+++ b/mesmer/mesmer_m/power_transformer.py
@@ -267,8 +267,10 @@ class PowerTransformerVariableLambda(PowerTransformer):
         for gridcell, lmbda in enumerate(lambdas.T):
             for year, y_lmbda in enumerate(lmbda):
                 with np.errstate(invalid="ignore"):  # hide NaN warnings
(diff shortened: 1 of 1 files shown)


mesmer/mesmer_m/power_transformer.py
```diff
@@ -200,13 +200,13 @@ def transform(self, monthly_residuals, yearly_T):
 
         transformed_monthly_resids = np.zeros_like(monthly_residuals)
 
-        # for i, lmbda in enumerate(lambdas.T):
-        #     for j,j_lmbda in enumerate(lmbda):
+        # for gridcell, lmbda in enumerate(lambdas.T):
+        #     for year,year_lmbda in enumerate(lmbda):
```
mathause, line 204:
```suggestion
        #     for year, year_lmbda in enumerate(lmbda):
```

mathause (approved):
Let's merge as is because this is definitely a plus. (However, that should eventually be rewritten (I don't like the "iterate over rows of the array approach" & this can maybe be vectorized).)

PR #448, opened 2024-05-16
clean up dependencies

### LINKED ISSUE #364: clean up dependencies
The dependencies are defined at several places. We should streamline them.

E.g. our environment adds gdal, but that is not directly needed; we need pyproj but that is never mentioned (it's a dependency of geopandas so it's in the stack but that not optimal); netcdf4 is missing, scipy is missing.

- https://github.com/MESMER-group/mesmer/blob/6f5d3e59dbc947273f28099cbc402ecef64f73dd/setup.cfg#L35
- https://github.com/MESMER-group/mesmer/blob/main/environment.yml
- https://github.com/conda-forge/mesmer-feedstock/blob/main/recipe/meta.yaml
- https://github.com/MESMER-group/mesmer/blob/main/docs/environment.yml

There are also a number of packages that are not required but usually needed - see #315

Related #175

---

This is something we absolutely should do, but it's more annoying to do than you'd guess...

mathause:
- [x] Closes  #364

This cleans up the dependencies in `setup.cfg` and `environment.yml` and explicitly lists all of our direct dependencies.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -38,6 +38,7 @@ Internal Changes
   By `Mathias Hauser`_.
 - Restore compatibility with pandas version v2.2 and xarray version v2024.02 (`#404 <https://github.com/MESMER-group/mesmer/pull/404>`_).
   By `Mathias Hauser`_.
+- Explicitely include all required dependencies (`#448 <https://github.com/MESMER-group/mesmer/pull/448>`_).
 
 
 v0.10.0 - 2024.01.04
--- a/ci/requirements/min-all-deps.yml
+++ b/ci/requirements/min-all-deps.yml
@@ -7,15 +7,17 @@ channels:
 dependencies:
  - python=3.9
  - cartopy
- - dask[complete]
- - gdal
+ - dask
+ - joblib
  - netcdf4
  - numpy
  - packaging
  - pandas=2.0
  - pooch
+ - pyproj
  - regionmask=0.9
  - scikit-learn
+ - scipy
  - shapely=1.8 # required by regionmask < 0.10
  - statsmodels=0.13
  - xarray=2023.04
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -30,7 +30,7 @@ Alternately, mesmer can be installed with pip:
 
 .. code-block:: bash
 
-   python -m pip install mesmer-emulator
+   python -m pip install mesmer-emulator[full]
 
 Development installation
 ------------------------
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -30,7 +30,7 @@ Alternately, mesmer can be installed with pip:
 
 .. code-block:: bash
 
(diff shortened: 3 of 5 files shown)


veni-vidi-vici-dormivi:
Okay, so do I understand that correctly that
- the `installation.rst` is only for documentation purposes and lists the bare necessities, but everything we use explicitly is also stated explicitly.  It doesn't contain version numbers for the packages because it is just for documentation.
- the `environment.yml` is for someone to build an environment to work with mesmer, which is why it also has cartopy and matplotlib etc. Would we want to add notebook infrastructure to that (ipykernel) - not yet I guess because we don't ship notebooks (yet). It also contains minimum version numbers
- and the setup.cfg is what I run when we push/release and what is installed on the user side when doing it via pip. It doesn't contain plotting utilities for example. 

Is there any way we don't have to change three files when changing dependencies, can we import them into one another?

mathause:
> * the `installation.rst` is only for documentation purposes and lists the bare necessities, but everything we use explicitly is also stated explicitly.  It doesn't contain version numbers for the packages because it is just for documentation.

Correct - we could add the min versions but I was too lazy for now

> * the `environment.yml` is for someone to build an environment to work with mesmer, which is why it also has cartopy and matplotlib etc. Would we want to add notebook infrastructure to that (ipykernel) - not yet I guess because we don't ship notebooks (yet). It also contains minimum version numbers

Correct. We also use it to set up the test environment.

> * and the `setup.cfg` is what I run when we push/release and what is installed on the user side when doing it via pip. It doesn't contain plotting utilities for example.

Correct this is what's installed when calling `pip install ...`. The test env get's defined in environment.yml file (although we also call `pip install -e .` so if a package was missing fom environment.yml it would get installed via pip). Yes let me add these optional dependencies to the extras. Unfortunately dask is transitive but not optional (as we use `open_mfdataset`).

> Is there any way we don't have to change three files when changing dependencies, can we import them into one another?

Not that I know. Actually it's even worse:
- I forgot to adapt ci/requirements/min-all-deps.yml (commit incoming)
- it also has to be defined for conda-forge (on the mesmer-feedstock [meta.yaml](https://github.com/conda-forge/mesmer-feedstock/blob/main/recipe/meta.
- (Actually, we should also have it in requirements.txt as long as we are using setup.cfg [but that is only for the benefit of github and becomes obsolete once we fully move to pyproject.toml])

However, the dependencies should not be such a moving target and are usually very stable. But you are right the dependency definition is a bit of a mess and more clean ups are certainly possible. (mesmer is somewhat weird in that you can install if from pypi and conda-forge but it's not really a package to be just used so far...)

mathause:
I think one issue we have is that the 3 environment.yml are in 3 different places:

```
./environment.yml
./ci/requirements/min-all-deps.yml
./docs/environment.yml
```

For xarray & regionmask the are all in one place (./ci/requirements) which helps a bit

veni-vidi-vici-dormivi (approved):

PR #451, opened 2024-05-16
deprecate mask_3D_frac_approx

mathause:
- [ ] Closes #xxx
 - [ ] Tests added

regionmask 0.12.0 offers it's own implementation of `mask_3D_frac_approx`, so at some point we can remove this function from mesmer. This takes the first step by marking it as deprecated.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -25,6 +25,9 @@ Breaking changes
 Deprecations
 ^^^^^^^^^^^^
 
+- Deprecated ``mask_3D_frac_approx`` as the functionality is now offered in regionmask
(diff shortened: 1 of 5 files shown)


veni-vidi-vici-dormivi:
Is that a function meant for the user? Wouldn't the user just use `mask_ocean_fraction`? Then we wouldn't need a depreciation.

veni-vidi-vici-dormivi (approved):

mathause:
> Is that a function meant for the user? Wouldn't the user just use `mask_ocean_fraction`? Then we wouldn't need a depreciation.

Yes, I did it for the off-chance that anyone would be using it. For which the chance is like `<= eps`, but I couldn't help myself :anguished:

Thanks for the review - aren't you taking the day off?

PR #452, opened 2024-05-16
use `mask_3D_frac_approx` from regionmask

mathause:
Uses the `mask_3D_frac_approx` function from regionmask, which should be faster than implementation we have here in mesmer. 

This PR builds #451 - we need to be careful to be merging into main!

### DIFF
--- a/mesmer/core/mask.py
+++ b/mesmer/core/mask.py
@@ -56,7 +56,7 @@ def mask_ocean_fraction(data, threshold, *, x_coords="lon", y_coords="lat"):
     land_110 = regionmask.defined_regions.natural_earth_v5_0_0.land_110
 
     try:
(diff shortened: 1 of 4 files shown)


veni-vidi-vici-dormivi (approved):

mathause:
Ok this was terrible but I learned some stuff :man_shrugging:

- basing a PR on another PR was a pain
  - the original diff was terrible, which I why I did this
  - however, this did not really help (changing my approach did)
- for the PR chain to work, the original PR (i.e. its branch) has to be on the upstream repo (I had to manually push the branch here)
- this was probably the reason why switching the base branch did not automatically work
- (the MESMER-X linting branch is fine as the mesmer_x branch is on the main repo; pushing to the original branch generally seems preferable [although that was not applicable here])
- I spent _waaay_ too much time on this

mathause:
At least the final diff looks nice :man_shrugging: 

Thanks for the review.

PR #456, opened 2024-05-23
Unshallow checkout

### LINKED ISSUE #454: unshallow git repo on rtd
While we are at it we should also unshallow the mesmer git repository on rtd (or fetch the tags). See: https://github.com/regionmask/regionmask/blob/2161e2c713fc8ae3c380264ddbeaaed68c7792ad/.readthedocs.yml#L10

Context: we display the version number in the top left corner of the docs. Which includes the version number, and if it's not a realease version, also the number of commits and it's hash. To compute the version number we need the git tag. However, as rtd does a shallow clone and then fetches the commit with a depth of 50, it only downloads the tag if we have fewer than 50 commits. So we just lost it and it now shows version 0

image

image

_Originally posted by @mathause in https://github.com/MESMER-group/mesmer/issues/453#issuecomment-2124937480_

veni-vidi-vici-dormivi:
- [x] Closes #454 

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -2,6 +2,9 @@ version: 2
 build:
   os: ubuntu-22.04
   jobs:
+    post_checkout:
+      # unshallow so the version works
+      - git fetch --unshallow
     pre_install:
         # see https://github.com/readthedocs/readthedocs.org/issues/8201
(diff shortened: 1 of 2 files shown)


mathause (approved):
Thanks

mathause:
The version works again

image

PR #457, opened 2024-05-24
rtd: use pip

### LINKED ISSUE #453: use pip to install env on readthedocs?
All dependencies of our doc build are pip-installable and we could therefore switch to installing the dependencies with pip, as discussed in #449. Using `python -m pip install .[docs]`, for the installation. This it's faster and we could remove _docs/environment.yml_. See regionmask/regionmask#450 for inspiration.

(my only fear is that we eventually want to convert notebooks in the docs build using nbconvert and have no good way to install a modern version of pandoc (a nbconvert dependency) on rtd)

mathause:
- [x] Closes #453

Test if the docs can be built using `python -m pip install -e .[docs]`. Potentially the docs need to be updated.


mathause:
This saves about 2 min from the docs build, so it's about twice as fast as before.

veni-vidi-vici-dormivi:
This looks more simple than before! I think that's great!

PR #458, opened 2024-05-24
Harmonic model bonus test

veni-vidi-vici-dormivi:
In this PR I add a more sophisticated test for the xarray harmonic model. I also deleted Shrutis test file because I think I already implemented all the ideas that were in that file.
 - [ ] Tests added


veni-vidi-vici-dormivi:
Doing this, I made two interesting observations:
1. for the sixth harmonic, the coefficients of the sine term will always end up at the first guess: 
$sin( \frac{\pi \cdot 6 \cdot mon}{6}) = sin(\pi * mon)$ 
with $mon \in [1, 12]$ as integers, this will always give 0 so there is no fit for the first two coefficients of this order. I think it is still fine to include the sixth harmonic in the fit because the cosine term can to some extent make up for missing the sine term. But we cannot test if coefficients are close for this order (in a straight forward way at least - still thinking about that). 

2. the tests I do here work fine if we have n_ts = 10 but for more time steps, the fitted orders increase to an extent where at 100 time steps, all the orders come out to be six which makes the fitting take muuuch longer. I am not sure why this happens yet. When fitting real data, it's fine. 

I'll work on this next week I guess.

tests/unit/test_harmonic_model.py
```diff
@@ -172,23 +172,48 @@ def test_fit_to_bic_numerical_stability():
...
-    "coefficients",
-    [
-        np.array([0, -1, 0, -2]),
-        np.array([1, 2, 3, 4, 5, 6, 7, 8]),
-        np.array([1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]),
-    ],
-)
-def test_fit_to_bic_xr(coefficients):
-    yearly_predictor = trend_data_2D(n_timesteps=10, n_lat=3, n_lon=2)
+def get_2D_coefficients(orders, n_lat=3, n_lon=2):
```
mathause, line 175: * Rename `orders` to `order_per_cell`? (or similar)

tests/unit/test_harmonic_model.py
```diff
@@ -172,23 +172,48 @@ def test_fit_to_bic_numerical_stability():
...
+    }
+
+    return xr.DataArray(data, dims=("cells", "coeff"), coords=coords, name="coeffs")
+
+
+def test_fit_to_bic_xr():
+    np.random.seed(0)
+    n_ts = 10
+
+    coefficients = get_2D_coefficients(orders=[1, 2, 3, 4, 5, 6], n_lat=3, n_lon=2)
```
mathause, line 207: Define orders as variable and use below in the `assert` check?

mathause (commented):
Nice - however, took me some time to wrap my head around. I added some suggestions.

tests/unit/test_harmonic_model.py
```diff
@@ -172,23 +172,48 @@ def test_fit_to_bic_numerical_stability():
...
+
+    LON, LAT = np.meshgrid(np.arange(n_lon), np.arange(n_lat))
+
+    coords = {
+        "coeff": coeff,
+        "lon": ("cells", LON.flatten()),
+        "lat": ("cells", LAT.flatten()),
+    }
+
+    return xr.DataArray(data, dims=("cells", "coeff"), coords=coords, name="coeffs")
```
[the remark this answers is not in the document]
veni-vidi-vici-dormivi: makes sense :sweat_smile

tests/unit/test_harmonic_model.py
```diff
@@ -172,23 +172,48 @@ def test_fit_to_bic_numerical_stability():
...
+    n_cells = n_lat * n_lon
+    coeff = np.arange(6 * 4)
+    max_order = 6
+
+    # generate decreasing coefficients with some noise
+    trend = np.repeat(np.linspace(1.2, 0.2, max_order), 4)
+    scale = np.tile([0, 5.0], (n_cells, max_order * 2))
+    rng = np.random.default_rng(0)
+    noise = rng.normal(loc=0, scale=0.1, size=(n_cells, max_order * 4))
+    data = trend * scale + noise
```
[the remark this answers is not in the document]
veni-vidi-vici-dormivi: Ah there is a mistake in there. So what I wanted to do here is to generate coefficients that are close to real ones. So I just checked how the coefficients come out when using the coarse data and this shows two things:
1. Generally, the coefficients decrease rather exponentially in magnitude towards higher order (I changed the linspace now to mimic that better). Which makes sense because the first order should capture the biggest variations and with each additional order it goes more into the details.
2. The every second coefficient is very close to zero which also makes sense because these are the coefficients responsible for scaling the seasonal cycle to the temperature:
$ T_{s,m,y} = \T_{s,y} + \sum_{i=1}^{n} c_{sin, 0} \cdot T_{s,y} + c_{sin, 1} \cdot sin(\frac{i \pi mon}{6}) + c_{cos, 0} \cdot T_{s,y} + c_{cos, 1} \cdot cos(\frac{i \pi mon}{6})$
so big $c_{term, 0} would mean that with rising temperatures the variability of the seasonal cycle changes a lot, but this is actually not the case. Instead every month has a rather constant offset from the mean temperature which is expressed through $c_{term, 1}.

So what I wanted to do here is to generate coefficients with a downwards sloping trend for higher orders, but make the first coefficients really small and the second ones large and add some variation to them so they are not all the same. `noise` is in this case no really a good term because this isn't actually noise but rather a way to generate different coefficients so they are not exactly the same for each order. Conceptually, there is cannot be noise in the coefficients I guess.

I'll add commentary here and implement the exponential decrease and rename noise.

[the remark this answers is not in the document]
mathause: Thanks! Very clear description.

tests/unit/test_harmonic_model.py
```diff
@@ -172,23 +172,51 @@ def test_fit_to_bic_numerical_stability():
...
-        np.array([1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]),
-    ],
-)
-def test_fit_to_bic_xr(coefficients):
-    yearly_predictor = trend_data_2D(n_timesteps=10, n_lat=3, n_lon=2)
+def get_2D_coefficients(order_per_cell, n_lat=3, n_lon=2):
+    n_cells = n_lat * n_lon
+    max_order = 6
+
+    # generate coefficients that reseble real ones
```
mathause, line 179:
```suggestion
    # generate coefficients that resemble real ones
```

tests/unit/test_harmonic_model.py
```diff
@@ -200,17 +228,27 @@ def test_fit_to_bic_xr(coefficients):
...
         vectorize=True,
         output_dtypes=[float],
-        kwargs={"coeffs": coefficients, "months": months},
+        kwargs={"months": months},
     )
 
+    # test if the model can recover the monthly target from perfect fourier series
     result = fit_to_bic_xr(yearly_predictor, monthly_target)
-
+    assert (result.n_sel.values == orders).all()
```
mathause, line 241:
```suggestion
    np.testing.assert_equal(result.n_sel.values, orders)
```

mathause (approved):

PR #460, opened 2024-05-29
Harmonic model return residuals

### LINKED ISSUE #459: Harmonic model testing
Increasing the number of years in the testing of the harmonic model fit in the following test

https://github.com/MESMER-group/mesmer/blob/dc60db6b1e154bde73f5c242a71792b458f41224/tests/unit/test_harmonic_model.py#L206-L241

fails. The selected orders increase even though the fit for the coefficients is close to zero. We should check the BIC and why it keeps fitting in this case. The fitting still works on real data, so I'm really not sure what is happening here. Even if it does not change the predictions significatntly, fitting up to the max order on every gridpoint massively slows down the fit.

Things we should check: 
- the BIC
- if it maybe works when the trend in the global temperature is less steep
- happy about suggestions...

veni-vidi-vici-dormivi:
- [ ] Closes #459
 - [x] Tests added

### DIFF
--- a/tests/unit/test_harmonic_model.py
+++ b/tests/unit/test_harmonic_model.py
@@ -53,12 +53,11 @@ def test_generate_fourier_series_np():
     ],
 )
 def test_fit_to_bic_np(coefficients):
+    # ensure original coeffs and order is estimated from fourier series without noise
     max_order = 6
     n_years = 100
     months = np.tile(np.arange(1, 13), n_years)
-    yearly_predictor = yearly_predictor = trend_data_1D(
-        n_timesteps=n_years, intercept=0, slope=1
-    ).values
+    yearly_predictor = trend_data_1D(n_timesteps=n_years, intercept=0, slope=1).values
     yearly_predictor = np.repeat(yearly_predictor, 12)
     monthly_target = generate_fourier_series_np(yearly_predictor, coefficients, months)
 
@@ -152,6 +151,7 @@ def test_fit_to_bic_xr():
     noisy_monthly_target = monthly_target + rng.normal(
         loc=0, scale=0.1, size=monthly_target.values.shape
     )
+    
     result = fit_to_bic_xr(yearly_predictor, noisy_monthly_target)
     xr.testing.assert_allclose(result["predictions"], monthly_target, atol=0.1)
     np.testing.assert_allclose(
--- a/tests/unit/test_harmonic_model.py
+++ b/tests/unit/test_harmonic_model.py
@@ -157,23 +157,25 @@ def test_fit_to_bic_xr():
     xr.testing.assert_allclose(result["predictions"], monthly_target, atol=0.1)
 
     # compare numerically one cell of one year
-    expected = np.array([
-                9.99630445,
-                9.98829217,
-                7.32212458,
-                2.73123514,
-                -2.53876124,
-                -7.07931947,
-                -9.69283667,
-                -9.6945128,
-                -7.08035255,
-                -2.53178204,
-                2.74790275,
-                7.34046832,
-            ]
-        )
+    expected = np.array(
+        [
+            9.99630445,
+            9.98829217,
+            7.32212458,
+            2.73123514,
+            -2.53876124,
+            -7.07931947,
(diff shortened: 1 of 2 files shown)


tests/unit/test_harmonic_model.py
```diff
@@ -52,23 +52,21 @@ def test_generate_fourier_series_np():
...
-    [np.repeat([-1, 1], 10 * 6), np.linspace(-1, 1, 10 * 12) * 10],
-)
-def test_fit_to_bic_np(coefficients, yearly_predictor):
+def test_fit_to_bic_np(coefficients):
     max_order = 6
-    months = np.tile(np.arange(1, 13), 10)
-
+    n_years = 100
+    months = np.tile(np.arange(1, 13), n_years)
+    yearly_predictor = yearly_predictor = trend_data_1D(
```
mathause, line 59:
```suggestion
    yearly_predictor = trend_data_1D(
```

tests/unit/test_harmonic_model.py
```diff
@@ -52,23 +52,21 @@ def test_generate_fourier_series_np():
         np.array([1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]),
     ],
 )
-@pytest.mark.parametrize(
-    "yearly_predictor",
-    [np.repeat([-1, 1], 10 * 6), np.linspace(-1, 1, 10 * 12) * 10],
-)
-def test_fit_to_bic_np(coefficients, yearly_predictor):
+def test_fit_to_bic_np(coefficients):
```
mathause, line 55:
```suggestion
def test_fit_to_bic_np(coefficients):
    # ensure original coeffs and ts is estimated from fourier series without noise
```

tests/unit/test_harmonic_model.py
```diff
@@ -239,15 +145,34 @@ def test_fit_to_bic_xr():
...
     rng = np.random.default_rng(0)
     noisy_monthly_target = monthly_target + rng.normal(
         loc=0, scale=0.1, size=monthly_target.values.shape
     )
     result = fit_to_bic_xr(yearly_predictor, noisy_monthly_target)
-    xr.testing.assert_allclose(result["predictions"], monthly_target, atol=0.2)
+    xr.testing.assert_allclose(result["predictions"], monthly_target, atol=0.1)
+    np.testing.assert_allclose(
+        result.predictions.isel(cells=0, time=slice(0, 12)).values,
+        np.array(
```
mathause, line 159: Can you move this out of the function?

mathause (approved):
Do you know if the tests are faster because they abort earlier or because the fitting is genuinely faster?

veni-vidi-vici-dormivi:
> Do you know if the tests are faster because they abort earlier or because the fitting is genuinely faster?

Should be because they are generally faster because also before they aborted at low orders (for few time steps).

tests/unit/test_harmonic_model.py
veni-vidi-vici-dormivi: Do you really mean time steps here? I would say we try for the order?

tests/unit/test_harmonic_model.py
veni-vidi-vici-dormivi: What do you mean? Write a new one, e.g. `test_fit_to_bic_xr_withnoise` ?

mathause:
Yes you are right it seems to be both :champagne:

tests/unit/test_harmonic_model.py
mathause: no ts -> timeseries; but yes that's not obvious - I updated the comment

tests/unit/test_harmonic_model.py
mathause: No, that you define a the array outside the `assert_allclose` function, e.g. `expected_predictions = np.array([...])`

veni-vidi-vici-dormivi: Of course, thanks 😅

mathause:
This brings down the time of the tests by 30 s (1/3) - another nice win :1st_place_medal:

PR #466, opened 2024-06-13
MESMER-X: first code updates

mathause:
A PR which does a first pass on the MESMER-X code.


yquilcaille (approved):
Everything ok with these changes. Code improved without hindering the functionalities.

PR #467, opened 2024-06-14
MESMER-X: decorator to supress warnings

mathause:
Created a decorator to suppress the warnings. This allows to avoid a level of indentation, unfortunately there is not much that is unwrapped. But should become slightly easier to read anyway. We absolutely need to be more specific which warnings we suppress, because currently we also miss deprecations and other warnings we want to see - but have to defer that until we have some running tests to see what warnings there are. github does not seem to get the correct diff...


mathause:
The diff works better if whitespace differences are ignored

image


merging

PR #474, opened 2024-06-27
Remove power transformer class

### LINKED ISSUE #462: test PowerTransformer against sklearn (for constant lambda)
Does not need to be super thorough but should ensure that the "trivial" case is the same as the reference implementation.

veni-vidi-vici-dormivi:
- [ ] Closes #462 
 - [ ] Tests added

In this PR I am removing the power transformer class in favor of the newly written xarray functions and adjusting the tests accordingly.


veni-vidi-vici-dormivi:
I now deleted the PowerTransformerVariableLambda class. Should we instead deprecate it properly? I thought it is unnecessary because because it was never properly implemented. What do you think @mathause?

tests/unit/test_power_transformer.py
```diff
@@ -66,110 +61,143 @@ def test_fit_power_transformer():
...
+    )
+    result = power_transformer._yeo_johnson_inverse_transform_np(transformed, lambdas)
 
     np.testing.assert_allclose(result, monthly_residuals, atol=1e-7)
 
 
-def test_standard_scaler():
-    # generate random data with mean different from 0 and std different from one
+def test_yeo_johnson_inverse_transform_np_sklearn():
+    # test if our inverse power trasform is the same as sklearns
```
mathause, line 146:
```suggestion
    # test if our inverse power trasform is the same as sklearn's for constant lambda
```

mathause (approved):
Good for me to just remove it. It also disentangles it as the `PoerTransformer` can also do a box-cox transformation which was however not adapted for variable $\lambda$.

veni-vidi-vici-dormivi:
I think the only point where we still need sklearn now is the tests. So do I see that correctly that we could now (again) have a separate environment for testing and for the users and exclude sklearn in the users environment, or not? Even if so, I would keep it at one environment at this point I think.

mathause:
It's always nice to have less dependencies but we can keep it for the moment. Maybe add a comment in setup.cfg?

PR #484, opened 2024-07-27
Move mesmer_m functionalities into stats module

veni-vidi-vici-dormivi:
I move the mesmer_m module `harmonic_model` and `power_transformer` into the `stats` module. This makes for shallower function calls. Along with this I did some renaming, privatizing of functions and nits for the documentation. I also adjusted the init files, the test and the api.rst.


mathause:
Nice - I'll wait for #480 before reviewing (but looks good).

veni-vidi-vici-dormivi:
Question: How do we want to display data types in the documentation? At the moment its a bit of a mess:
image

Do we them to just appear cursive or in one of the two styles seen above in purple font with the option for boldness? The above is achieved with this code:

```python
"""fit an auto regression

    Parameters
    ----------
    data : xr.DataArray
        A ``xr.DataArray`` to estimate the auto regression over.
    dim : str
        Dimension along which to fit the auto regression.
    lags : int
        The number of lags to include in the model.

    Returns
    -------
    :obj:`xr.Dataset`
        Dataset containing the estimated parameters of the ``intercept``, the AR
        ``coeffs``, the ``variance`` of the residuals and the number of observations ``nobs``.
    """
   ```
   
The cursive for the types of the parameters is also adopted by numpy. xarray uses the bold option with the frame but there these are at the same time links to the documentation of these types so since we do not use such links, this could be confusing. Also, the purple framed but not bold option is just the same style as code bits which can be kind of ambiguous as in this case where we also use this style for variables. I think in the end it's a question about aesthetics but it would be nice to unify it already a bit here.

**Comment by mathause** [comment]
> Question: How do we want to display data types in the documentation?

Erm, from the visuals I prefer the framed option but it's more work. Do you have an opinion? Also happy to defer to later.

**Comment by veni-vidi-vici-dormivi** [comment]
Yeah me too. This is how I did it now for the power transformer functions. Feel free to have a look at the documentation. Then we can keep it like that here and keep it in mind for the rest for later.

File: docs/source/api.rst [hunk 1]
```diff
@@ -32,8 +32,31 @@ Auto regression
...
+   ~stats.draw_auto_regression_monthly
+
+Harmonic Model
+--------------
+
+.. autosummary::
+   :toctree: generated/
+
+   ~stats.generate_fourier_series
+   ~stats.fit_harmonic_model
```
mathause, line 47: Should we align the name of these two functions? (Also fine to not do it or defer to later).

mesmer/stats/_auto_regression.py
```diff
@@ -618,6 +618,10 @@ def fit_auto_regression_monthly(monthly_data, time_dim="time"):
     where :math:`\\tau \\in \\{1, \\ldots, N\\}` counts the seasons of some seasonal cycle, here the
     months of a year :math:`(N=12)` and :math:`t` counts the repetitions of this seasonal cycle,
     here the years. Here :math:`\\epsilon` is a white noise process, i.e. :math:`\\epsilon \\sim N(0, \\sigma^2)`.
+    The covariance matrix of the driving white noise process should be estimated on the residuals of the AR(1)
+    process. The residuals are returned here and should be passed to
+    :func:`find_localized_empirical_covariance_monthly <mesmer.stats.find_localized_empirical_covariance_monthly>`.
```
mathause, line 623: I didn't know that you can do that

mesmer/stats/_harmonic_model.py
```diff
@@ -16,12 +16,13 @@
 
 
 def _generate_fourier_series_np(yearly_predictor, coeffs, months):
-    """construct a Fourier Series
+    """construct a Fourier Series from the yearly predictor with given coeffs.
+    The order of the Fourier series is inferred from the size of the coeffs array.
```
mathause, line 20:
```suggestion

    The order of the Fourier series is inferred from the size of the coeffs array.
```

mesmer/stats/_harmonic_model.py
```diff
@@ -103,17 +105,18 @@ def generate_fourier_series(yearly_predictor, coeffs, time, time_dim="time"):
     return predictions.transpose(time_dim, ...)
 
 
-def fit_fourier_series_np(yearly_predictor, monthly_target, first_guess):
-    """execute fitting of the harmonic model/fourier series
+def _fit_fourier_coeffs_np(yearly_predictor, monthly_target, first_guess):
+    """fit the coefficients of a Fourier Series to the data using least squares for the
+    given order of the Fourier Series, infered from the size of the `first_guess` array.
```
mathause, line 110:
```suggestion
    given order, which is inferred from the size of the `first_guess` array.
```

mesmer/stats/_harmonic_model.py
```diff
@@ -182,21 +185,19 @@ def calculate_bic(n_samples, order, mse):
...
-    # NOTE: removed -2 now because for order=0 the first two coefficients dissapear as sin(0) == 0
-    # removed this now because we no longer fit beta0 and beta1
-
     return n_samples * np.log(mse) + n_params * np.log(n_samples)
 
 
-def fit_to_bic_np(yearly_predictor, monthly_target, max_order):
-    """choose order of Fourier Series to fit for by minimising BIC score
+def _fit_fourier_order_np(yearly_predictor, monthly_target, max_order):
+    """choose order of Fourier Series to fit for by minimising BIC score.
```
mathause, line 192:
```suggestion
    """determine order of Fourier Series for by minimizing BIC score.
```

mesmer/stats/_harmonic_model.py
```diff
@@ -182,21 +185,19 @@ def calculate_bic(n_samples, order, mse):
...
+    For each order, the coefficients are fit using least squares internally.
```
mathause, line 193:
```suggestion
    For each order, the coefficients are fit using least squares.
```

mesmer/stats/_power_transformer.py
```diff
@@ -65,21 +81,25 @@ def _yeo_johnson_transform_np(data, lambdas):
 
 
 def _yeo_johnson_inverse_transform_np(data, lambdas):
-    """invert residuals using Yeo-Johnson transformation with variable lambda
+    """invert data using the inverse Yeo-Johnson transformation with variable lambda.
```
mathause, line 84:
```suggestion
    """inverse Yeo-Johnson transformation with variable lambda.
```

mathause (approved):
Looks good - nice to have that in the stats module!

mesmer/stats/_auto_regression.py
veni-vidi-vici-dormivi: hihihi 😊😊😊 I love it, this might be happening a lot now 😂

docs/source/api.rst
veni-vidi-vici-dormivi: Hm good point. So reading the paper again "harmonic model" summarizes the two parts (1) the linear regression of the local yearly temperature and (2) the Fourier series. So `generate_fourier_series` should actually be `generate_harmonic_model` I guess because it contains both. However, since we don't actually fit the linear regression but only the coefficients of the Fourier series (thus `fit_fourier_coeffs_np`) it would probably be cleanest to separate these two things so that it is more obvious which is what... I'm not happy with the overall structure anyways.

I think it would be maybe most intuitive to have `fit_harmonic_model` call `fit_fourier_order` using the monthly deviations from the yearly temperature and this one contain `fit_fourier_coeffs` and `generate_fourier_series. Then add the predictions to the yearly temperature at the very end in `fit_harmonic_model`. But then, if we ever want to reintroduce the fitting of the linear regression we would have think it all over again.

mathause: About your last point as long as we do $X \rightarrow X_m$ this is given. If we ever want to do $X \rightarrow Y_m$, it needs re-thinking. But I think that chance for this is very small.

veni-vidi-vici-dormivi: Good point. I think if it comes to that the whole approach needs a thorough review, so much so that it's okay that there are no provisions for it in the code.

mathause (commented):
I do like your updates but should we do them in separate PR's? Until the "refactoring of months argument in harmonic model" commit there were no functional changes. (But I also understand if that is too cumbersome...)

I think it is easier to do the functional changes after merging this PR first (but need to get the commits into other branches).



<details>

This was my first idea but I think you can do it even easier

```bash

# create new branch with state _before_ the "refactoring of months argument in harmonic model" commit
git co -b hm_no_months 587cbb2c5fa86cec15ffc44d6885bd4091458b41

# cherry pick "refactoring of months argument in harmonic model" commit
git cherry-pick 66f9926409ca40dcb120a2acf9c68e58a3bb1aa8

# DO THE SAME WITH THE "factor out linear regression harmonic model" ... "fix test" commits


git switch m_goes_stats 
# undo changes on m_goes_stats branch
git reset 587cbb2c5fa86cec15ffc44d6885bd4091458b41 --soft

git push origin HEAD --force

# merge PR, update local main, merge main into hm_no_months branch, merge this PR
```

</details>


```
# create new branch with "refactoring of months argument in harmonic model" commit
git co -b hm_no_months 66f9926409ca40dcb120a2acf9c68e58a3bb1aa8

# create new branch with last commit ("fix tests")
git co -b hm_no_regression bc85cee7378c3b76cdab7fa7870d7fd8c0a23205

git switch m_goes_stats 
# undo changes on m_goes_stats branch before_ the "refactoring of months argument in harmonic model" commit
git reset 587cbb2c5fa86cec15ffc44d6885bd4091458b41 --soft

git push origin HEAD --force

# merge this PR, update local main, merge main into hm_no_months branch
# open new PR on github merge this PR, and repeat
```

veni-vidi-vici-dormivi:
Okay all done and clean. I will open the other two PR now and reference some of the comments made here.

PR #487, opened 2024-08-07
refactor creation of month array

veni-vidi-vici-dormivi:
We can move the creation of the month array used for the fourier series to the deepest level right before the computation of the fourier series itself.


mathause (approved):
Thanks!

Issue #56, opened 2021-05-31
auto-check headers


mathause:
See znicholls/mesmer-openscmrunner#6

mathause:
We could probably use ruff flake8-copyright for this. See

- https://docs.astral.sh/ruff/rules/#flake8-copyright-cpy
- https://docs.astral.sh/ruff/settings/#flake8-copyright

Would depend on #343

PR #490, opened 2024-08-08
use ruff for code linting

### LINKED ISSUE #343: use ruff as linter?
Much faster than the current isort + flake8 approach (note: black needs to stay<sup>*</sup>). -integrates several linters at the same time. See `[tool.ruff]` section in https://github.com/pydata/xarray/blob/main/pyproject.toml for an example configuration. Will also need to update our documentation.

<sup>*</sup> for the moment - ruff has a autoformat rule that black 24 reversed. So we should wait for the update.

mathause:
Add ruff configuration. I don't necessarily want to switch but we could... 

- [x] closes #343

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -2,5 +2,5 @@
 
  - [ ] Closes #xxx
  - [ ] Tests added
- - [ ] Passes `isort . && black . && flake8`
+ - [ ] Passes `ruff check . && black .`
  - [ ] Fully documented, including `CHANGELOG.rst`
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -2,5 +2,4 @@
 
  - [ ] Closes #xxx
  - [ ] Tests added
- - [ ] Passes `ruff check . && black .`
  - [ ] Fully documented, including `CHANGELOG.rst`
--- a/Makefile
+++ b/Makefile
@@ -45,21 +45,16 @@ help:
 
 .PHONY: format
 format: $(VENV_DIR)  ## auto-format the code using relevant tools
-	make isort
+	make ruff
 	make black
-	make flake8
 
 .PHONY: black
 black: $(VENV_DIR)  ## auto-format the code using black
 	$(VENV_DIR)/bin/black $(FILES_TO_FORMAT_PYTHON) docs/source/conf.py
 
-.PHONY: flake8
+.PHONY: ruff
 flake8: $(VENV_DIR)  ## lint the code using flake8
-	$(VENV_DIR)/bin/flake8 $(FILES_TO_FORMAT_PYTHON)
-
-.PHONY: isort
-isort: $(VENV_DIR)  ## lint the code using flake8
-	$(VENV_DIR)/bin/isort $(FILES_TO_FORMAT_PYTHON)
+	$(VENV_DIR)/bin/ruff $(FILES_TO_FORMAT_PYTHON)
 
 .PHONY: docs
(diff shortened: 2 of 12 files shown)


veni-vidi-vici-dormivi:
Would it replace sort and flake8? The changes it made are nice. In #343 you said we would need to update our documentation, why? Also that we should wait for an update so that it does not interfere with black, is that out now?

mathause:
Right - I made an issue on that :flushed: the formatting tools are mentioned in https://github.com/MESMER-group/mesmer/blob/main/docs/source/development.rst#formatting so they'd need updating (and also the Makefile, but we could also delete it), and it's less important now that we use pre-commit CI anyways.

The formatting issue is still open - astral-sh/ruff#9745

mathause:
And yes it would replace flake8 and isort (and we could consider adding more rules, see e.g. https://github.com/shapely/shapely/blob/775f7792ecd945fa30b3f98b46d99f996e275f00/pyproject.toml#L81 for an inspiration (or also the ruff docs).

pyproject.toml
```diff
@@ -8,3 +8,28 @@ build-backend = "setuptools.build_meta"
...
+# E731: do not assign a lambda expression, use a def
+ignore = [
+  "E402",
+  "E501",
+  "E731",
+]
+select = [
+  "F", # Pyflakes
+  "E", # Pycodestyle
+  "W", # warnings
```
mathause, line 29:
```suggestion
  "E", # pycodestyle - Errors
  "W", # pycodestyle - Warnings
```

mathause:
Ok, switched to ruff entirely. I keep the isort and flake config around - otherwise we can get a mess if isort or flake is ever called. One issue with this is that there are formatting checks in ruff - so they might trigger an error that is fixed by black. Before we could run the formatting checks after the formater. If that is a problem we can turn off the formatting checks: https://github.com/astral-sh/ruff-pre-commit/issues/74#issuecomment-2003961490

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -2,5 +2,5 @@
 
  - [ ] Closes #xxx
  - [ ] Tests added
- - [ ] Passes `isort . && black . && flake8`
+ - [ ] Passes `ruff check . && black .`
```
veni-vidi-vici-dormivi, line 5: I feel like it is kind of ambiguous to have a check for that here when you can't merge without it anyways... What do you think about removing it?

Makefile
```diff
@@ -45,21 +45,16 @@ help:
...
 	make black
-	make flake8
 
 .PHONY: black
 black: $(VENV_DIR)  ## auto-format the code using black
 	$(VENV_DIR)/bin/black $(FILES_TO_FORMAT_PYTHON) docs/source/conf.py
 
-.PHONY: flake8
+.PHONY: ruff
 flake8: $(VENV_DIR)  ## lint the code using flake8
```
veni-vidi-vici-dormivi, line 56: is the comment still right?

veni-vidi-vici-dormivi (approved):

veni-vidi-vici-dormivi:
And think about what happens to #343 and #56 now.

Makefile
mathause: no - thanks

.github/PULL_REQUEST_TEMPLATE.md
mathause: I still prefer to do the linting locally but yes we can remove this

mathause (commented):
Thanks for the review. I will update the two files. We can close #343 but I leave #56 for another PR.

We should maybe remove the Makefile - I don't actually use it and haven't checked if it still works (I mean I guess it does but still).

PR #491, opened 2024-08-08
add mesmer_m example script

veni-vidi-vici-dormivi:
Here an example script that executes the mesmer_m workflow for the coarse test dataset.


examples/example_mesmer_m.py
```diff
@@ -0,0 +1,172 @@
...
+        geodist, localisation_radii=LOCALISATION_RADII
+    )
+
+    weights = xr.ones_like(AR1_fit.residuals.isel(gridcell=0))
+    weights.name = "weights"
+
+    localized_ecov = mesmer.stats.find_localized_empirical_covariance_monthly(
+        AR1_fit.residuals, weights, phi_gc_localizer, "time", 30
+    )
+
```
mathause, line 121: could we split the training and creating the emulations into two functions? To emphasize that normally you would save the params and then load them.

mathause: maybe also a `load_data` function

examples/example_mesmer_m.py
```diff
@@ -0,0 +1,172 @@
...
+    # make emulations
+    nr_emus = 10
+    buffer = 20
+    seed = 0
+    ref_period = slice("1850", "1900")
+
+    # preprocess tas
+    ref = tas_y.sel(time=ref_period).mean("time", keep_attrs=True)
+    tas_y = tas_y - ref
+    tas_stacked_y = mask_and_stack(tas_y, threshold_land=1 / 3)
```
mathause, line 131: wasn't that already done above? or is this on purpose? (which is also fine)

examples/example_mesmer_m.py
```diff
@@ -0,0 +1,172 @@
...
+    )
+
+    weights = xr.ones_like(AR1_fit.residuals.isel(gridcell=0))
+    weights.name = "weights"
+
+    localized_ecov = mesmer.stats.find_localized_empirical_covariance_monthly(
+        AR1_fit.residuals, weights, phi_gc_localizer, "time", 30
+    )
+
+    # make emulations
```
mathause, line 122: add a comment that we create emulations from the model temperature - and it could also be an option to go Tglob -> T local annual -> T local monthly?

Or is this inconsistent with the other level of comments?

mathause (commented):
Thanks - the mesmer-m approach is now really concise and well readable. I am happy to merge as a script but I would eventually like to create notebooks that we can include in our documentation (ah but it's still 40s (on my laptop) which is long to run for a doc build).

- do you know if Shruti ever fits more than one ensemble member and/ or scenario? Because this may not be possible in the current form :grimacing:
- would also be good to have an integration test from this script (but again the 40 s is very long for this)

mathause:
I quickly timed the individual parts (to have a reference)

what | duration
---  | ---:
data loading | 2s
mesmer.stats.fit_harmonic_model | 9s
mesmer.stats.fit_yeo_johnson_transform | 15s
mesmer.stats.find_localized_empirical_covariance_monthly | 7s
emulations | 2s

veni-vidi-vici-dormivi:
> Thanks - the mesmer-m approach is now really concise and well readable. 

Thank you, I think so too, we did a good job 😊 

> I am happy to merge as a script but I would eventually like to create notebooks that we can include in our documentation (ah but it's still 40s (on my laptop) which is long to run for a doc build).

Hm yeah a notebook for the docs would be really cool. I would merge as is now and pretty it up later when we are done with the numerical issues. 

> do you know if Shruti ever fits more than one ensemble member and/ or scenario? Because this may not be possible in the current form 😬

I thought about that already and in the paper she writes that she uses only one scenario. If there are several members she splits them into a training and a test set and for some there are several members in the test set. I could not infer from the paper if the model is trained on several members together or for each member individually. The notebook provided for the paper definitely has no functionality to handle several members so the aggregation happened either before that or training happened separately and the evaluation statistics were then aggregated. We should ask her about it.

> would also be good to have an integration test from this script (but again the 40 s is very long for this)

Hm we could try and do it with less years maybe. Interesting that the transformer is the slowest part, good to know. And cool to see that the emulation is very quick, that's nice.

examples/example_mesmer_m.py
veni-vidi-vici-dormivi: Yes it was, it was kind of to illustrate that this part would be done independently from the training above but I see that that is not very intuitive. Splitting it as you said above would be helpful.

examples/example_mesmer_m.py
veni-vidi-vici-dormivi: No I think that's a good point.

veni-vidi-vici-dormivi:
Okay I switched to a notebook now. It has lots of room for more commenting and some visualisations but we can do that later.

mathause:
Thanks. Can you add two more titles: "Calibrate model" title, and "Find localized empirical covariance" and remove the inputs ("Clear all inputs")

mathause:
I will add some more comments to the notebook - I will updat it directly, so please don't do the same ;-)

mathause (approved):
Thanks - I approve but maybe you can read through quickly as well.

veni-vidi-vici-dormivi:
> but maybe you can read through quickly as well

Through what?

mathause:
> > but maybe you can read through quickly as well
> 
> Through what?

:open_mouth: oh I forgot to push my changes

mathause:
Ah no I pushed to `origin` instead of your fork

```diff
-git push origin HEAD
+git push vici HEAD
```

veni-vidi-vici-dormivi:
I added some small details and some more info on the time coordinate, that we save. Actually that's only necessary when you want/need the time coordinate on the emulations to be the same as for the data but if that is not necessary one could also create one oneself.

mathause:
Thanks!

PR #493, opened 2024-08-09
Implement eigh() fallback

### LINKED ISSUE #413: Reconsider try Cholesky approach
We now use Cholesky decomposition for finding the localization radius and for drawing innovations (see #408). When the covariance matrix is not positive definite we use `eigh()`. Depending on how often this is actually the case we could switch to using Cholesky and if it fails, the user needs to provide an alternative method. 

Another add on: localizing the covariance matrix ensures positive semi-definiteness (see [#402](https://github.com/MESMER-group/mesmer/issues/402#issuecomment-1981054915) for an explanation) but not positive definiteness which is needed for Cholesky. We could check already in `_localized_covariance` if cov is positive definite and set the decomposition method there. We could implement several options here for the user:
1) have the option for the user to enforce positive definiteness of the adjusted covariance matrix throw an error if not. However, this is not overly useful I think, because what is the user gonna do then?
2) warn if the adjusted covariance is not positive definite and set decomposition method for drawing emulations to `eigh()` or some user given option 
3) in any case, give the option to chose the decomposition method when drawing emulations in `_auto_regression`.

In combination with this it maybe would make sense to completely factor out all covariance wrangling into its own module... but then nothing is really happening in `_localized covariance`... Maybe the best option would be to make a `covariance_utils` module to be used in `_localized_covariance` and _auto_regression. In `_auto_regression` also some refactoring could be done, see #165.

### LINKED ISSUE #420: Raise LinAlg error in try Cholesky
https://github.com/MESMER-group/mesmer/blob/903535a9a43fff05c81bff49e44700f4dbc25281/mesmer/stats/_auto_regression.py#L496C13-L496C18

This line is not covered by tests. I wrote it because I thought there might be other possible linAlg errors than a non positive definite matrix but I think that is actually not true, see here: https://github.com/numpy/numpy/blob/d35cd07ea997f033b2d89d349734c61f5de54b0d/numpy/linalg/linalg.py#L772C5-L772C67
Nevertheless, they write in the documentation "Raises: LinAlgError -If the decomposition fails, **for example**, if a is not positive-definite."
I tried to come up with a patch but failed. However, this might become obsolete when we implement #413. So not investing more time in this for now.

### LINKED ISSUE #616: Bug in `_minimize_local_discrete`
I mentioned this already in #413 but then forgot about it. Consider the following case in `_minimize_local_discrete`:
The very first localized covariance matrix is singular and thus returns a negative loglikelyhood of `inf`. 
Then `current_min` is still `inf`. `res` is also `inf`, a warning gets raised but we continue. `res < current_min` in this case becomes `float("inf") < float("inf")` and yields `False`. Thus we land at `sel =  i - 1` and `sel = -1`. We return sequence[-1]`, in this case the last localization radius, without ever checking its likelihood or if the resulting localized covariance matrix is not singular. In my opinion, this is a bug and we should fix it before releasing MESMER v1.0. 

It can be fixed by actually skipping results that are `inf` (rather complicated) or raising an Error if the first element already results is `inf` if we think that once the localization radii return singular matrices the rest will also always return singular matrices. This is inuitively true when the radii are ordered (which we do not explicitly require) and higher radii lead to less localization and thus to "too little regularized" matrices. At the moment, we always stop when `res == inf` and return the element from the iteration before (NOTE: I think the comment in the code is misleading).

https://github.com/MESMER-group/mesmer/blob/bd011cf4da1ead8c3419a16ba875bd59ff484bf6/mesmer/core/utils.py#L54-L80

Also check the following test cases:

```python
@pytest.mark.parametrize(
    "values, expected",
    [((5, 4, 3, 2, 3, 0), 3), ((1, 0, 1, 2), 1), ((float("inf"), 1, 2, 3), 1),
     ((3, 2, 1, float("inf"), 0), 4), ((float("inf"), float("inf"), float("inf")), 2),
     ((3, float("inf"), 2, 1, float("inf"), 2), 3)],
)
def test_minimize_local_discrete(values, expected):

    data_dict = {key: value for key, value in enumerate(values)}

    def func(i):
        return data_dict[i]

    result = mesmer.core.utils._minimize_local_discrete(func, data_dict.keys())

    assert result == expected
```

veni-vidi-vici-dormivi:
Implement eigh() as a fallback for loglikelihood estimation in covariance localization. Since physically, the covariance matrix can be positive semidefinite we should allow singular true. I also implemented some bug fixes mentioned in #413.

 - [x] Closes #413
 - [x] Closes #420
 - [x] Closes #616
 - [x] Tests added


veni-vidi-vici-dormivi:
We could cherry pick the rng commit to another PR, I just opened the branch for all thing covariance related as a tryout in the beginning.

veni-vidi-vici-dormivi:
Before I fix the rest of the tests, could you look at it @mathause. I don't think it's pretty what I did to `mesmer.utils.minimize_local_discrete` but I couldn't think of a better way. maybe it would also be okay to check cholesky for every localization radius if it is rather likely that one where the matrix is pos def is the better one in which case we would only do the estimation with eigh once or max twice.

mesmer/stats/_localized_covariance.py
```diff
@@ -265,7 +269,9 @@ def _find_localized_empirical_covariance_np(data, weights, localizer, k_folds):
     return localization_radius, covariance, localized_covariance
 
 
-def _ecov_crossvalidation(localization_radius, *, data, weights, localizer, k_folds):
+def _ecov_crossvalidation(
```
mathause, line 272: Alternatively we could turn this into a class and save the `method` as state. Also not super nice, but we can avoid the change to `_minimize_local_discrete` (I think that should work...).


```python
class _EcovCrossvalidation:

    def __init__(self, method=None):

       self.method = method or "cholesky"

    def crossvalidate(localization_radius, *, data, weights, localizer, k_folds):

            ...

            try:
                # sum log likelihood of all crossvalidation folds
                nll += _get_neg_loglikelihood(data_cv, localized_cov, weights_cv, self.method)
            except np.linalg.LinAlgError:
                # NOTE: this error is thrown by np.linalg.cholesky not by the logpdf anymore
                warnings.warn(
                    f"Singular matrix for localization_radius of {localization_radius}."
                    "\n Switching to eigh().",
                    LinAlgWarning,
                )
                # switch to eigh from now on
                self.method = "eigh"
                nll += _get_neg_loglikelihood(data_cv, localized_cov, weights_cv, self.method)

                ...

```

mesmer/stats/_localized_covariance.py
```diff
@@ -291,19 +297,22 @@ def _ecov_crossvalidation(localization_radius, *, data, weights, localizer, k_fo
 
         try:
             # sum log likelihood of all crossvalidation folds
-            nll += _get_neg_loglikelihood(data_cv, localized_cov, weights_cv)
+            nll += _get_neg_loglikelihood(data_cv, localized_cov, weights_cv, method)
         except np.linalg.LinAlgError:
+            # NOTE: this error is thrown by np.linalg.cholesky not by the logpdf anymore
```
mathause, line 302: In case `method == "eigh"` ever raises an error it just happend in the lower statement - is that ok or do we need

```python
if method == "eigh":
    raise e
```

here?

mesmer/core/utils.py
```diff
@@ -72,6 +75,9 @@ def _minimize_local_discrete(func, sequence, **kwargs):
             sel = i - 1
             if sel == 0:
                 warnings.warn("First element is local minimum.", OptimizeWarning)
+            elif sel == -1:
+                raise ValueError("First radius already inf.")
+
```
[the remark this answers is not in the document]
veni-vidi-vici-dormivi: I don't think that we ever get there because when the result is inf we terminate the loop. Because inf is always $\geq$`current_min` and thus we return, no?

mesmer/stats/_localized_covariance.py
mathause: The third alternative is to use an inner function, but that's probably even less readable (defining inner functions is pretty common but updating state is not and requires `nonlocal`)

```python
def outer():
    state = 5
    def inner():
        nonlocal state
        print(state)
        state = 7
    return inner

func = outer()
func()
func()
```

veni-vidi-vici-dormivi: I would prefer the class.

mesmer/stats/_localized_covariance.py
veni-vidi-vici-dormivi: Ugh yes, but I cannot come up with a test for it... very unsatisfying.

veni-vidi-vici-dormivi:
Note: diff works better with ignore whitespace

mathause (approved):
Thanks that looks good and clean. One last thing we could do (and I leave it up to you): define `_ecov_crossvalidation = _EcovCrossvalidation()` and rename `crossvalidate` to `__call__`, then we can call `_ecov_crossvalidation(...)` again as before (but then this function is purely internal...)

veni-vidi-vici-dormivi:
Stumbled upon #187. Just wanted to mention it here, but it seems like it does not necessarily speak against allowing singular true?

mathause:
I think the problem in #187 was that certain folds did not contribute to the nll. We should make sure this doesn't happen. Need to take a look at the code...

veni-vidi-vici-dormivi:
Another not so nice side effect about all of this is that if we select a singular covariance matrix and use eigh for drawing samples this could lead to slight differences in the emulations between machines...

veni-vidi-vici-dormivi:
Maybe [this](https://en.wikipedia.org/wiki/Ostrich_algorithm) is the solution to go for. Also because an ostrich looks very similar to an emu.

mathause:
I've avoided reviewing this because I am unsure what's best (different results on different OSs is annoying but prob. unavoidable, however, if #187 happens it would be bad). Should/ can we wait for #521 so we can easily play with different models / scenarios before deciding here? (I think there _were_ models with singular matrices for Lea, maybe only in CMIP5.)

veni-vidi-vici-dormivi:
Okay

veni-vidi-vici-dormivi:
Actually it seems that allow singular is ignored if cov is a scipy.stats.Covariance object so we should not pass it to the scipy function, but only use it for our internal decision to switch to eigh()

PR #494, opened 2024-08-12
more stable yeo johnson roundtrips

mathause:
- [x] Addresses the issue mentioned in https://github.com/MESMER-group/mesmer/pull/476#issuecomment-2198048690
 - [x] Tests added

With this we can take another stab at #476. It also seems a bit faster. Modeled after scikit-learn/scikit-learn#26188

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -90,6 +90,8 @@ Yeo-Johnson power transformer
       `#455 <https://github.com/MESMER-group/mesmer/pull/455>`_)
 -  adjust the first guess to assume the data is normally distributed (
    `#429 <https://github.com/MESMER-group/mesmer/pull/429>`_)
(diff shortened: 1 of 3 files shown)


veni-vidi-vici-dormivi:
Huh interesting, so that means that `np.power` yields less precision than `np.exp` and `np.log`, right? That's good to know.

veni-vidi-vici-dormivi (approved):

mathause:
Thanks. I think its more the `np.expm1` and `np.log1p`. From the docstrings:

* `expm1`: This function provides greater precision than ``exp(x) - 1`` for small values of ``x``.
* `log1p`: For real-valued input, `log1p` is accurate also for `x` so small that `1 + x == 1` in floating-point accuracy.

(interesting the names are not symmetric...)

veni-vidi-vici-dormivi:
Interesting. Thanks for doing that!

PR #495, opened 2024-08-14
Switch to numpy rng for seeding

### LINKED ISSUE #312: use 'modern' random number generator machinery
We currently use `np.random.seed` and `np.random.multivariate_normal` which is discouraged by numpy. We should switch to the modern approach. See details in https://numpy.org/doc/stable/reference/random/index.html#module-numpy.random

To be considered together with #35, I realize this was already mentioned in #168

veni-vidi-vici-dormivi:
Switch from `np.random.seed()` to `np.random.default_rng()` for drawing emulations

 - [x] Closes #312 
 - [x] Tests added


mesmer/stats/_auto_regression.py
```diff
@@ -879,16 +881,13 @@ def _draw_auto_regression_monthly_np(
...
-    np.random.seed(seed)
-
     # draw innovations for each month
     innovations = np.zeros([n_samples, n_ts // 12 + buffer, 12, n_gridcells])
     if covariance is not None:
         for month in range(12):
             cov_month = covariance[month, :, :]
             innovations[:, :, month, :] = _draw_innovations_correlated_np(
-                cov_month, n_gridcells, n_samples, n_ts // 12, buffer
+                cov_month, seed, n_gridcells, n_samples, n_ts // 12, buffer
```
mathause, line 890: That is a problem here, because `default_rng` will always be re-initialized, so we end up with the same random numbers for each month.

mathause (commented):
I think we need to pass `rng` to `_draw_innovations_correlated_np`. Maybe add a small test to ensure the random numbers are not the same when the function is called twice (without re-initializing rng) and is the same when doing so?

mesmer/stats/_auto_regression.py
veni-vidi-vici-dormivi: Ah I see, thanks!

tests/unit/test_auto_regression.py
```diff
@@ -694,6 +694,38 @@ def test_draw_auto_regression_monthly_np_buffer(buffer):
...
+    )
+
+    np.testing.assert_equal(res, res2)
+
+    # ensure that innovations for each month are different though
+    # (even with the same covariance matrix)
+    jan = res[:, 0::12, :]
+    feb = res[:, 1::12, :]
+
+    np.testing.assert_raises(AssertionError, np.testing.assert_equal, jan, feb)
```
mathause, line 726:
```suggestion
    pytest.raises(AssertionError):
        np.testing.assert_equal(jan, feb)
```

Or maybe

```suggestion
    assert np.not_equal(jan, feb).all() # or .any()
```

mathause: looking at it again I like the second option better

tests/unit/test_auto_regression.py
```diff
@@ -694,6 +694,38 @@ def test_draw_auto_regression_monthly_np_buffer(buffer):
...
+    )
+
+    np.testing.assert_equal(res, res2)
+
+    # ensure that innovations for each month are different though
+    # (even with the same covariance matrix)
+    jan = res[:, 0::12, :]
+    feb = res[:, 1::12, :]
+
+    assert np.not_equal(jan, feb).any()
```
mathause, line 726:
```suggestion
    assert np.not_equal(jan, feb).all()
```

mathause (approved):
Thanks!

tests/unit/test_auto_regression.py
veni-vidi-vici-dormivi: I'm not sure if there is not a tiny chance that actually a value happens to be the same in both...

veni-vidi-vici-dormivi: Yes but I am thinking if maybe in the future the functionality changes so that for some reason there are new numbers (for example we change the method of drawing to numpy again or something different) then this test should not indicate that something went wrong...

mathause: Fair enough - add a comment instead?

PR #496, opened 2024-08-14
yeo-johnson: performance optimization

mathause:
- [ ] Closes #xxx
 - [ ] Tests added

This reduces the computation of `mesmer.stats.fit_yeo_johnson_transform` to below 10s. Still not blazingly fast but a bit faster with an acceptable increase in complexity (IMO). #494 helped as I could see that all four cases require `np.log1p(np.abs(data))` (which is constant), so we can pre-compute that in one go (instead doing it every iteration with semi-costly indexing shenanigans). Similarly, it seems faster to create intermediate arrays instead of indexing them twice (`lmbds`).

I did not optimize `_yeo_johnson_inverse_transform_np` because it is not used in the inner loop of the optimization and thus much less critical.

I hope I got all signs correct :man_shrugging:

### DIFF
--- a/mesmer/stats/_power_transformer.py
+++ b/mesmer/stats/_power_transformer.py
@@ -89,10 +89,10 @@ def _yeo_johnson_transform_split(data):
         transf[sel_a] = data_log1p[sel_a]
 
         lmbds = lambdas[sel_b]
-        transf[sel_b] = (np.expm1(data_log1p[sel_b] * lmbds)) / lmbds
+        transf[sel_b] = np.expm1(data_log1p[sel_b] * lmbds) / lmbds
 
         lmbds = 2 - lambdas[sel_c]
-        transf[sel_c] = -(np.expm1(data_log1p[sel_c] * lmbds)) / lmbds
+        transf[sel_c] = -np.expm1(data_log1p[sel_c] * lmbds) / lmbds
 
         transf[sel_d] = -data_log1p[sel_d]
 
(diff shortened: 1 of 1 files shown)


mathause:
I don't do `data_log1p = np.sign(data) * np.log1p(np.abs(data))` because it's cheaper and case d happens less often than case c...

mesmer/stats/_power_transformer.py
```diff
@@ -134,22 +153,23 @@ def _yeo_johnson_optimize_lambda_np(monthly_residuals, yearly_pred):
     monthly_residuals = monthly_residuals[~isnan]
     yearly_pred = yearly_pred[~isnan]
 
+    # initialize constant arrays
+    _yeo_johnson_transform = _yeo_johnson_transform_optimized(monthly_residuals)
```
veni-vidi-vici-dormivi, line 157: ~~I don't quite understand, when you do it like this, the variables initialized here are not initialized again when we call it in the loop afterwards, and just `inner` is recomputed? How cool is that?~~ Oh now I get it, it returns `inner`... And the constant arrays are being stored in the function definition itself? That is very cool.

veni-vidi-vici-dormivi (approved):

veni-vidi-vici-dormivi:
Wow I have to say that is pretty cool. It is more complex but also the speed up is quite impressive. Thanks! 

The signs should all be correct when all tests pass. 🤸‍♀️

mesmer/stats/_power_transformer.py
mathause: Yes, the variables are "bound" to the function.

PR #501, opened 2024-08-20
Add MESMER-M integration tests

veni-vidi-vici-dormivi:
- [x] Tests added


mathause:
Thanks - can you save the coeffs as netCDF and not pickle (I know it's a bit more annoying but we want to move away from pickle)

mathause:
I haven't looked into it so not sure what is going on but a bit strange that `none` of the environments get it right. I thought this does not have to be high priority but then it would be good to stabilize the numerics for memser-m. What helped for the covariance thingy was testing the code on cfc vs exo, which have different cpus with different sets of SIMD instructions. But I am not really working today - if I find time I'll take a look tomorrow or else after our retreat...

veni-vidi-vici-dormivi:
The problem is that the fit of the harmonic mode is not the same on different machines. Therein I cannot narrow it down anymore but changing switching to a more exact computation of the Jacobin brought down the amount of mismatched elements significantly, to only two. I don't think there is any. problem in the power transformer, it most likely just gives different results because it fits on the residuals of the harmonic model.

veni-vidi-vici-dormivi:
Uff okay I have to structure this, here's a list of what I tried and what changed.

- `scipy.optimize.least_squares`
  - method 'lm': Max absolute difference among violations: 1.3
  - defaults to 'trf' (and jac 2-point): Max absolute difference among violations: 0.4
  - method = 'trf' and jac = 'cs': Max absolute difference among violations: 0.007 and that for only 2 elements (I looked at the elements and they do not appear special to me. It is one gridcell that looks pretty normal but a rather high order)
  - trying different losses (the ones that are applicable), solvers, and tolerances to the fit didn't change that much.

- `scipy.optimize.curve_fit`
  - defaults to method 'lm': Max absolute difference: 1.5**e-06**, this calls `leastsq` and not `least_squares`.
 
There seems to me no obvious reason which method to choose, since I think that neither our Jacobian is likely rank deficient nor sparse and we don't set any bounds. I don't really know what to do here.

veni-vidi-vici-dormivi:
Okay I think (for the moment) this is good enough. I can open an issue on this after merging so we can come back to it maybe in the future. Also, should I separate the changes in the harmonic model and the power transformer from the integration tests into (3) separate PRs or just rename this PR to "MESMER-M integration test stabilizing"?

mesmer/stats/_harmonic_model.py
```diff
@@ -135,14 +135,16 @@ def _fit_fourier_coeffs_np(yearly_predictor, monthly_target, first_guess):
 
     """
 
-    def func(coeffs, yearly_predictor, mon_target):
-        return _generate_fourier_series_np(yearly_predictor, coeffs) - mon_target
+    def residuals_from_fourier_series(coeffs, yearly_predictor, mon_target):
+        return mon_target - _generate_fourier_series_np(yearly_predictor, coeffs)
```
mathause, line 139: Conceptually I prefer the other way round (subtract the reference) - but fine to keep.

mesmer/stats/_harmonic_model.py
```diff
@@ -135,14 +135,16 @@ def _fit_fourier_coeffs_np(yearly_predictor, monthly_target, first_guess):
...
 
+    # Use curve_fit to optimize the coefficients
```
mathause, line 141: Is this comment still applicable?

tests/integration/test_calibrate_mesmer_m.py
```diff
@@ -0,0 +1,147 @@
...
+    )
+
+    weights = xr.ones_like(AR1_fit.residuals.isel(gridcell=0))
+    weights.name = "weights"
+
+    localized_ecov = mesmer.stats.find_localized_empirical_covariance_monthly(
+        AR1_fit.residuals, weights, phi_gc_localizer, "time", 30
+    )
+
+    # merge into one dataset
```
mathause, line 119: I am not a huge fan of this there and back renaming and merging. Would it be bad to save each Dataset as individual netCDF? Or merge into a datatree? (But maybe later).

tests/unit/test_harmonic_model.py
```diff
@@ -81,11 +82,11 @@ def test_fit_fourier_order_np(coefficients):
     yearly_predictor = np.repeat(yearly_predictor, 12)
     monthly_target = _generate_fourier_series_np(yearly_predictor, coefficients)
 
-    selected_order, estimated_coefficients, predictions = _fit_fourier_order_np(
+    _, estimated_coefficients, predictions = _fit_fourier_order_np(
         yearly_predictor, monthly_target, max_order=max_order
     )
 
-    np.testing.assert_equal(selected_order, int(len(coefficients) / 4))
+    # np.testing.assert_equal(selected_order, int(len(coefficients) / 4))
```
mathause, line 89: Can you enable these tests again?

tests/integration/test_draw_mesmer_m.py
```diff
@@ -0,0 +1,104 @@
...
+    local_variability_inverted = mesmer.stats.inverse_yeo_johnson_transform(
+        local_variability_transformed,
+        pt_coeffs,
+        tas_stacked_y.tas,
+    )
+
+    # add the local variability to the monthly harmonic
+    result = monthly_harmonic_emu + local_variability_inverted.inverted
+    result = result.to_dataset(name="tas")
+
```
mathause, line 87:
```suggestion

    test_file = TEST_PATH / "test_mesmer_m_realisations_expected.nc"

```

tests/integration/test_draw_mesmer_m.py
```diff
@@ -0,0 +1,104 @@
...
+    # save
+    if update_expected_files:
+        result.to_netcdf(TEST_PATH / "test_mesmer_m_realisations_expected.nc")
```
mathause, line 90:
```suggestion
        result.to_netcdf(test_file)
```

tests/integration/test_draw_mesmer_m.py
```diff
@@ -0,0 +1,104 @@
...
+        pytest.skip("Updated emulations.")
+
+    # testing
+    else:
+        exp = xr.open_dataset(
+            TEST_PATH / "test_mesmer_m_realisations_expected.nc", use_cftime=True
+        )
```
mathause, line 97:
```suggestion
        exp = xr.open_dataset(test_file, use_cftime=True)
```

tests/integration/test_calibrate_mesmer_m.py
```diff
@@ -0,0 +1,147 @@
...
+    else:
+        # load expected values
+        expected_params = xr.open_dataset(
+            TEST_PATH / "test-mesmer_m-params.nc", use_cftime=True
+        )
+        # np.testing.assert_allclose(expected_params.hm_coeffs.values, calibrated_params.hm_coeffs.values, atol=1e-7)
+        # np.testing.assert_allclose(
+        #     expected_params.pt_xi0.values, calibrated_params.pt_xi0.values
+        # )
+        xr.testing.assert_allclose(expected_params, calibrated_params, rtol=1e-2)
```
mathause, line 147: I think `rtol=1-e2` does not reflect the overall precision - can we tighten this (for some of them) by splitting it into the individual components again?

mathause (commented):
Thanks - that's very good to have. I add some first suggestions - the biggest being to not merge the coeffs - let me know if you don't think that's a good idea.

mathause:
Oh and could you merge main - unfortunately you will run into some merge conflicts from #508 and #512

tests/unit/test_harmonic_model.py
veni-vidi-vici-dormivi: Yess, after your performance optimization I can, thanks a lot! This was about to give me real headache.

tests/integration/test_calibrate_mesmer_m.py
veni-vidi-vici-dormivi: I like the idea of having one dataset with all the calibrated parameters per ESM. I did some renaming in the actual modules such that everything can be merged without problems. What do you think about that?

mathause:
Do you happen to know if the estimates are stable for each OS? Could we do os-dependent tolerances? (Not asking you to sink more time into this PR atm...)

veni-vidi-vici-dormivi:
> Do you happen to know if the estimates are stable for each OS? Could we do os-dependent tolerances? (Not asking you to sink more time into this PR atm...)

Sorry for never replying to this, yes the estimates are different from OS to OS, I remember I needed to increase the tolerance somewhere for windows to also pass. I guess we could do OS dependent tolerances but it's kind of awkward, since the tolerance then depends on which OS you produced the data with originally. Like now with the data from macOS, windows would need a larger tolerance, but if we ever push a dataset that was generated with windows it would have to be the other way around... so I'm not sure. but I like #563, although it is going back to what we did in #402, but oh well.

mathause:
> since the tolerance then depends on which OS you produced

Good point! Let's not do this.

mathause:
`lambda_coeffs` and `hm_coeffs` both have `coeff` as dimension and consequently the shorther one is broadcast to the size of the bigger. In this case `lambda_coeffs` has 24 entries along the coeff dimension of which only 2 are used.

veni-vidi-vici-dormivi:
Yes I agree that that's not so nice and I also am not that convinced of having one parameter file per ESM anymore, so it's fine for me to have a file or netcdf group? 👀👀 for each statistical method. However, I still think it is good to have the lambda parameters also along a coeff dimension in case we or someone else ever implements different lambda functions.

veni-vidi-vici-dormivi:
Ah okay we had that already, just without coordinates. Yeah I would keep the coordinates but we can revert the unique naming

mathause:
Ok so we have _very_ large differences some of the lambda_coeffs because the model wants very small lambdas and its `~ 1 / coeffs[0]`. Let us check how different the lambdas themselves are.

mathause:
The lambdas are `allclose` to `atol=1e-4`. So that's 'acceptable'.

Looking a bit closer at the coeffs: the largest ones are `~1e13` with a difference of `1e12` (comparing results from the laptop of @veni-vidi-vici-dormivi and me), however, they lead to the _same_ $\mathcal{L}$ (on my machine) - thus no wonder they can be so different. These `coeffs` leads to lambdas of `~1e-14` (so not quite `< eps` but close).

I also had a look at `minimize` with `"Nelder-Mead"` - the function has a default `tol` of `1e-4` and - if I understand this correctly - it says it converged if either the `delta_x` or `f(delta_x)` between two iterations is `< tol`. In practice it seems to abort the optimization loop for a $\Delta \mathcal{L}$ of `~1e-7`.

I played it through with one set of example data. Comparing `coeffs = [1e7, 0.1]` with `coeffs = [2e7, 0.1]` leads to a $\Delta \mathcal{L}$ of `~1e-7` and a $\Delta \lambda$ of `~1e-7` which leads to a difference in transformed data of also about `1e-7`. (Note a _larger_ `coeffs[0]` leads to a _smaller_ $\lambda$).

Ok what does this all mean? Good question :man_shrugging: I think we have two options

1. Add an upper bound to `coeff[0]` because at values of `1e8` (?) there is no longer a meaningful difference between them - given they lead to the same $\mathcal{L}$ and almost the same $\lambda$ and transformed data. The choosen upper bound is arbitrary - my suggestion is `1e10` - between the `1e12` where $\Delta\mathcal{L}=0$ and the `1e-7` which is the lower bound (given `tol`) and it's a nice number :shrug: I think Shruti also had an upper bound (of 1 which does not make much sense?)?
2. Use a saner function than the logistic regression (i.e. a linear regression).

---

@veni-vidi-vici-dormivi Can I ask you to pull this branch again, set the bounds to `bounds = np.array([[0, 1e10], [-0.1, 0.1]])`, re-generate the calibrated params and upload them again? If you want to separate the output datasets that would also be fine with me :wink: (but not necessary). I ask you because I want to play around with this more and then I directly have params that were generated on a different machine.

veni-vidi-vici-dormivi:
I like 1. and it makes sense to me. Thanks for pinning this down!

mathause:
Yes let's see how we can set `atol` and `rtol` for the `lambda_coeffs`...

veni-vidi-vici-dormivi:
> It's annoying that this is so brittle...

Yeah a linear function would solve that... But I would still want to defer it to #517 to implement a linear one and then use it for testing...

veni-vidi-vici-dormivi:
Wuhuu 🎉

mathause:
:rocket:

What did we say about the saved params? Do we take them apart again?

veni-vidi-vici-dormivi:
...yes. I'll do it later today. If you approve I can merge after that.

tests/integration/test_calibrate_mesmer_m.py
```diff
@@ -0,0 +1,222 @@
...
+
+        # the tols are set to the best we can do
+        # NOTE: it is always rather few values that are off
+        np.testing.assert_allclose(
+            expected_hm_params.coeffs,
+            harmonic_model_fit.coeffs,
+            atol=2e-5,
+        )
+        np.testing.assert_allclose(
+            expected_pt_params.coeffs, pt_coefficients.coeffs, atol=1e-4
```
mathause, line 207:
```suggestion
            expected_pt_params.coeffs, pt_coefficients.coeffs, atol=1.5e-4
```

veni-vidi-vici-dormivi:
good way to avoid pulling 😆

veni-vidi-vici-dormivi:
🥲

tests/integration/test_calibrate_mesmer_m.py
```diff
@@ -0,0 +1,222 @@
...
+
+        # the tols are set to the best we can do
+        # NOTE: it is always rather few values that are off
+        np.testing.assert_allclose(
+            expected_hm_params.coeffs,
+            harmonic_model_fit.coeffs,
+            atol=2e-5,
+        )
+        np.testing.assert_allclose(
+            expected_pt_params.coeffs, pt_coefficients.coeffs, atol=1.5e-4
```
mathause, line 207:
```suggestion
            expected_pt_params.coeffs, pt_coefficients.coeffs, atol=1.e-4, rtol=1.5e-4
```

mathause:
:sweat_smile:

mesmer/stats/_power_transformer.py
```diff
@@ -203,15 +207,15 @@ def get_lambdas_from_covariates(coeffs, yearly_pred):
...
+    if not isinstance(coeffs, xr.DataArray):
+        raise TypeError(f"Expected a `xr.DataArray`, got {type(coeffs)}")
 
     if not isinstance(yearly_pred, xr.DataArray):
         raise TypeError(f"Expected a `xr.DataArray`, got {type(yearly_pred)}")
 
     lambdas = xr.apply_ufunc(
         lambda_function,
-        coeffs.lambda_coeffs,
+        coeffs,
```
mathause, line 218: Last question: what's the reason you renamed those and now require a `DataArray`?

mathause (approved):
Ok I think we can finally finish this - thanks for all your work!

mesmer/stats/_power_transformer.py
mathause: I never know :man_shrugging: ok let's go with the `DataArray` but rename back to `lambda_coeffs`.

veni-vidi-vici-dormivi: that's what I prepared 😁

PR #503, opened 2024-08-20
fix minimum versions of dependencies

### LINKED ISSUE #175: add minimum supported versions
We should explicitly define which versions of our dependencies are minimally supported. We should then set up tests for an environment using these dependencies. The background is given in [NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html).

- Example of dependencies requirements: [regionmask/py37-min-all-deps.yml](https://github.com/regionmask/regionmask/blob/main/ci/requirements/py37-min-all-deps.yml)
- Checking if the given min deps is compatible with NEP-29 [xarray/min_deps_check.py](https://github.com/pydata/xarray/blob/main/ci/min_deps_check.py)
- github actions [regionmask/ci-additional.yaml](https://github.com/regionmask/regionmask/blob/main/.github/workflows/ci-additional.yaml)

However, this is is difficult because of the pickle files only conditionally backward compatible.

mathause:
- [x] Closes #175

Defines minumum versions for all dependencies. Ensures they follow NEP-29 (according to the xarray interpretation). Also updates the current lower bounds. (I decided to use the newest allowed version for all (direct) dependencies. It is of course always allowed to support older versions, but I don't see less upside for this as I once did.)

B.T.W. no need to review the min_deps_check.py script.


veni-vidi-vici-dormivi:
Why an additional ci yml?

mathause:
Ultimately because that is how it's done in xarray... Probably because it's not strictly code testing, but I can move it over.

veni-vidi-vici-dormivi (approved):

mathause:
Thanks - I hope this does not bite you in #501

Issue #492, opened 2024-08-08
better first guess for `fit_yeo_johnson_transform`


mathause:
We can speed up the `fit_yeo_johnson_transform` by passing a better first guess, assuming the trend is 0. We can get the first guess using:

```python
from sklearn.preprocessing import PowerTransformer

l = PowerTransformer().fit(tas_stacked_y.tas).lambdas_

# we can calculate xi_0 from lambda as
xi_0 = (2 - l) / l
```

veni-vidi-vici-dormivi:
Hm but instead of `tas_stacked_y.tas` with would use `resids_after_hm.tas[month]` right? So the assumption would be that there is a skew of the monthly residuals w.r.t. to the yearly values but that it is constant and not dependent on the yearly temperature value. That's a good idea. But we would need to do it 12 times too. Does that pay off?

mathause:
> Hm but instead of `tas_stacked_y.tas` with would use `resids_after_hm.tas[month]` right?

Yes

> Does that pay off?

The idea is that there is not much trend and that it's much faster to fit one param than 2 - that starting at a good point for $\xi_0$ speeds up the minimization. It helps, but only by about 10% - so much less than I would have hoped.

mathause:
I could try again with much lower precision for the first guess - most of the iterations are spent honing in the estimate. The fit uses `sp.optimize.brent` with a tolerance of about `1e-8`. For our purpose `1e-2` is probably enough.

Only problem: the `tol` param is not exposed in `PowerTransformer().fit`.

https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.brent.html

PR #512, opened 2024-08-30
fourier series: more consistent with standard definition

mathause:
- [ ] Closes #xxx
 - [ ] Tests added

Make our fourier series more consistent with the standard usage (e.g. on wikipedia for the fourier series or in numpy for fft<sup>1</sup>), [defined as](https://en.wikipedia.org/wiki/Fourier_series#math_Eq.2):

$$s_N(x) = A_0 + \sum_{n=1}^N \left(A_n \cos \left(2 \pi \tfrac{n}{P} x \right) + B_n \sin \left(2 \pi \tfrac{n}{P} x  \right) \right)$$

So this PR does two things:
1. switch the order of $\cos$ and $\sin$ so that $A_n$ is associated with $\cos$
2. use fractional years from $0...\frac{11}{12}$ instead of $\frac{1}{12}...1$. Note that this is nothing more than a phase shift with an angle of -30°. This phase shift leads to different values for $A_n$ and $B_n$ but to the same generated series.

<sup>1</sup> well fft uses the exponential form but that's equivalent

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -128,10 +128,11 @@ Harmonic model
 - de-duplicate the expression of months in their harmonic form (`#415 <https://github.com/MESMER-group/mesmer/pull/415>`_)
   move creation of the month array to the deepest level (`#487 <https://github.com/MESMER-group/mesmer/pull/487>`_).
 - fix indexing of harmonic model coefficients (`#415 <https://github.com/MESMER-group/mesmer/pull/415>`_)
--  Refactor variable names, small code improvements, fixes and clean docstring (
-   `#415 <https://github.com/MESMER-group/mesmer/pull/415>`_,
-   `#424 <https://github.com/MESMER-group/mesmer/pull/424>`_, and
-   `#433 <https://github.com/MESMER-group/mesmer/pull/433>`_)
+-  Refactor variable names, small code improvements, fixes and clean docstring
+   (`#415 <https://github.com/MESMER-group/mesmer/pull/415>`_,
+   `#424 <https://github.com/MESMER-group/mesmer/pull/424>`_,
+   `#433 <https://github.com/MESMER-group/mesmer/pull/433>`_, and
+   `#512 <https://github.com/MESMER-group/mesmer/pull/512>`_)
 - add tests (
   `#431 <https://github.com/MESMER-group/mesmer/pull/431>`_, and
   `#458 <https://github.com/MESMER-group/mesmer/pull/458>`_)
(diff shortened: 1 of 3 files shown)


mathause:
To mention the motivation with this - if we disregard the covariates we can use fft to estimate the coefficients with is _much_ faster. With this change the coeffs are much easier to compare:

```python
import numpy as np

import mesmer

# create test data
n_years = 5
n_months = n_years * 12

months = np.tile(np.arange(12), n_years)
alpha = 2 * np.pi * months / 12

monthly_target = 3.14 * np.cos(alpha) + 2.72 * np.sin(alpha) + np.random.randn(n_months)

# use mesmer to estimate coeffs w/o covariates
yearly_predictor = np.zeros(n_months)

coeffs, mse = mesmer.stats._harmonic_model._fit_fourier_coeffs_np(
    yearly_predictor,
    monthly_target,
    first_guess=np.zeros(4)
)
print(coeffs)

# use fft to estimate coeffs
freq = np.fft.rfftfreq(n_years * 12, 1/12)
sel = freq == 1.

coeffs_rfft = np.fft.rfft(monthly_target) / n_months

a_1 = 2 * coeffs_rfft[sel].real
b_1 = -2 * coeffs_rfft[sel].imag

# compare them
np.testing.assert_allclose(coeffs[1], a_1, rtol=1e-5)
np.testing.assert_allclose(coeffs[3], b_1, rtol=1e-5)


```

veni-vidi-vici-dormivi:
Okay, but can we disregard the covariates?

mathause:
> Okay, but _can_ we disregard the covariates?

I don't know<sup>1</sup>. It is a bit academic but helped me to understand the whole thing better. We could use fft to compute the `first_guess` potentially for all frequencies at once (as the fft is - well - _fast_, in contrast to my idea in #492). Although I am not sure if covariates for the first frequency influences the others.

For me this is a bit cleaner in any case & not very costly as it's not released in this form. 

Actually, I just wondered if we should reorganize coeffs even a bit more?

 state| $a_0$ | $a_1$ | $b_0$ | $b_1$
--- | --- | --- | --- | ---
current      | $\sin \cdot \mathrm{T}$ | $\sin$ | $\cos \cdot \mathrm{T}$ | $\cos$
this PR      | $\cos \cdot \mathrm{T}$ | $\cos$ | $\sin \cdot \mathrm{T}$ | $\sin$
even better? | $\cos$ | $\sin$ | $\cos \cdot \mathrm{T}$ | $\sin \cdot \mathrm{T}$

---

<sup>1</sup>We could find out if the covariates can be disregarded, also computing the BIC or deviance with and without the covariates.

mathause:
Ok let's keep the order of coeffs as suggested here (cosT, cos, sinT, sin). Is the PR ok for you, @veni-vidi-vici-dormivi?

veni-vidi-vici-dormivi (approved):

veni-vidi-vici-dormivi:
Yes definitely thank you.

mathause:
I thought if there was a better order of the coeffs for point 4 in #519 - but I did not come to any conclusion. So I suggest we merge as is and leave that for the future (if ever...).

PR #513, opened 2024-08-30
drop python 3.9 support

mathause:
- [ ] Closes #xxx
 - [ ] Tests added

[SPEC 0](https://scientific-python.org/specs/spec-0000) suggests we can drop python 3.9. So let's do that.

1. I removed the CI for 3.11 on purpose - I think testing 3.10 and 3.12 is enough (I never really had a failure due to the python version).
2. I did not know that a new syntax is supported for `isinstance` - interesting
   ```diff
   - isinstance(3, (int, str))
   + isinstance(3, int | str)
   ```

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -22,11 +22,11 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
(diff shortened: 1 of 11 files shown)


veni-vidi-vici-dormivi (approved):
Cool thanks!

PR #516, opened 2024-09-03
performance optimize `_generate_fourier_series_np`

### LINKED ISSUE #410: optimize `_generate_fourier_series_np`

mathause:
- [x] Closes #410
 - [ ] Tests added

Performance optimization of `_generate_fourier_series_np`. Builds on #512 so the diff is actually smaller. This does two main things:

1. `cos(alpha)` is periodic - so we only need to compute it once for each month (instead of `n_years * 12` times). We then reshape the `yearly_predictor` to `n_years x 12` such that broadcasting ensures each element is multiplied with the coefficients
2. Change the order of operation: sum the coefficients before multiplying - this avoids a lot of multiplication and sum operations. So instead of
   $$\sum\left((a_i \cdot T + b_i) \cdot \cos(\alpha) + (c_i \cdot T + d_i) \cdot \sin(\alpha)\right)$$
   we do
   $$\sum\left(a_i \cdot \cos(\alpha) + c_i \cdot \sin(\alpha)\right) \cdot T + \sum\left(b_i \cdot \cos(\alpha) + d_i \cdot \sin(\alpha)\right)$$

This speeds up `_generate_fourier_series_np` about 9 times and `fit_harmonic_model` about 4 times (on my machine). We do pay this with a _small_ loss in precision (see changes in tests)

---

What we do additionally:
1. Combine the $\cos$ and $\sin$ array into one so we can directly multiply them with the coeffs
2. use `@` to compute the `product-sum`
3. optimize the creation of the `alpha` array for the cost of some readability

What else could we do:
1. Split the function into two to pre-allocate the `cos_sin` array similar #496
2. use `fft` to get better inital conditions (https://github.com/MESMER-group/mesmer/pull/512#issuecomment-2321208656)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -128,10 +128,11 @@ Harmonic model
 - de-duplicate the expression of months in their harmonic form (`#415 <https://github.com/MESMER-group/mesmer/pull/415>`_)
   move creation of the month array to the deepest level (`#487 <https://github.com/MESMER-group/mesmer/pull/487>`_).
 - fix indexing of harmonic model coefficients (`#415 <https://github.com/MESMER-group/mesmer/pull/415>`_)
--  Refactor variable names, small code improvements, fixes and clean docstring (
-   `#415 <https://github.com/MESMER-group/mesmer/pull/415>`_,
-   `#424 <https://github.com/MESMER-group/mesmer/pull/424>`_, and
-   `#433 <https://github.com/MESMER-group/mesmer/pull/433>`_)
+-  Refactor variable names, small code improvements, fixes and clean docstring
+   (`#415 <https://github.com/MESMER-group/mesmer/pull/415>`_,
+   `#424 <https://github.com/MESMER-group/mesmer/pull/424>`_,
+   `#433 <https://github.com/MESMER-group/mesmer/pull/433>`_, and
+   `#512 <https://github.com/MESMER-group/mesmer/pull/512>`_)
 - add tests (
   `#431 <https://github.com/MESMER-group/mesmer/pull/431>`_, and
   `#458 <https://github.com/MESMER-group/mesmer/pull/458>`_)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -147,11 +147,12 @@ Harmonic model
 - de-duplicate the expression of months in their harmonic form (`#415 <https://github.com/MESMER-group/mesmer/pull/415>`_)
   move creation of the month array to the deepest level (`#487 <https://github.com/MESMER-group/mesmer/pull/487>`_).
 - fix indexing of harmonic model coefficients (`#415 <https://github.com/MESMER-group/mesmer/pull/415>`_)
--  Refactor variable names, small code improvements, fixes and clean docstring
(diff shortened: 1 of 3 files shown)


mesmer/stats/_harmonic_model.py
```diff
@@ -35,21 +35,33 @@ def _generate_fourier_series_np(yearly_predictor, coeffs):
...
+    order = coeffs.size // 4
+
+    # create 2D array of angles with shape (months, order)
+    # as 2 * np.pi * k * np.arange(12).reshape(-1, 1) / 12 but faster
+
+    factor = 2 * np.pi / 12
+    k = np.arange(1.0, order + 1)
+    alpha = np.arange(12 * factor, step=factor).reshape(-1, 1) * k
+
+    yearly_predictor = yearly_predictor.reshape(-1, 12)
```
veni-vidi-vici-dormivi, line 51: Would move this to the bottom before it is actually used in `seasonal_cycle =`

CHANGELOG.rst
```diff
@@ -147,10 +147,11 @@ Harmonic model
 - de-duplicate the expression of months in their harmonic form (`#415 <https://github.com/MESMER-group/mesmer/pull/415>`_)
   move creation of the month array to the deepest level (`#487 <https://github.com/MESMER-group/mesmer/pull/487>`_).
 - fix indexing of harmonic model coefficients (`#415 <https://github.com/MESMER-group/mesmer/pull/415>`_)
--  Refactor variable names, small code improvements, fixes and clean docstring (
-   `#415 <https://github.com/MESMER-group/mesmer/pull/415>`_,
-   `#424 <https://github.com/MESMER-group/mesmer/pull/424>`_, and
-   `#433 <https://github.com/MESMER-group/mesmer/pull/433>`_)
+-  Refactor variable names, small code improvements, fixes and clean docstring
```
veni-vidi-vici-dormivi, line 150:
```suggestion
-  Refactor variable names, small code improvements, optimization, fixes and clean docstring
```

veni-vidi-vici-dormivi (approved):

mathause:
Thanks for the review. Let's merge.

Issue #517, opened 2024-09-04
allow different lambda functions (besides logistic)?


mathause:
We have a TODO on allowing other lambda functions:

https://github.com/MESMER-group/mesmer/blob/8f8c8a06d299423997d9010617f734f830c497d4/mesmer/stats/_power_transformer.py#L256

E.g. logistic, constant, linear. I think we could maybe do this by collecting the necessary info in a class for each covariate function:

```python
class LogisticLambda:
    
    bounds = np.array([[0, np.inf], [-0.1, 0.1]])
    first_guess = np.array([1, 0])

    @staticmethod
    def lambdas_from_covariates(coeffs, covariate):
        return 2 / (1 + coeffs[0] * np.exp(covariate * coeffs[1]))

class ConstantLambda:
    
    bounds = None
    first_guess = np.array([1])

    @staticmethod
    def lambdas_from_covariates(coeffs, covariate):
        return np.full_like(covariate, fill_value=coeffs)

def _get_lambda_cls(function):

    if function == "logistic":
        return LogisticLambda()
    elif function == "constant":
        return ConstantLambda()
    else:
        raise ValueError()

...

```

```diff
- lambdas = lambda_function(coeffs, yearly_pred)
+ lambda_cls = _get_lambda_cls("constant")
+ lambdas = lambda_cls.lambdas_from_covariates(coeffs, yearly_pred)
```

---

As a side note: we have similar goals and challenges as for MESMER-X: extending standard statistical methods with covariates...

mathause:
I am still considering this for v1. But in contrast to my suggestion above I would turn the all power transformer functions into a class. This makes switching out the covariate function almost trivial (no need to pass the chosen covariate to the individual function).

```python
class YeoJohnson:

    def __new__(cls, covariate_function):
        if covariate_function == "linear":
            return YeoJohnsonLinear

    def lamda_function(): raise NotImplemented()

    def fit(): ...

    def transform(): ...

    def inverse_transform(): ...

class YeoJohnsonLinear(YeoJohnson):

    def lamda_function(): ...

YeoJohnson("linear")
# YeoJohnsonLinear
```

This would also remove power-transformer related functions from `mesmer.stats` (`lambda_function`, `get_lambdas_from_covariates`, `fit_yeo_johnson_transform`, `yeo_johnson_transform`, and `inverse_yeo_johnson_transform`)

PR #518, opened 2024-09-04
power transformer: coeffs as array

mathause:
- [ ] Closes #xxx
 - [ ] Tests added

This would be a step towards #517 - gather the coeffs in an array instead of different variables. This should make it agnostic to the number of coeffs there are.

What I am not happy is the naming (as always) - the DataArray is now named `coeffs` and it has a dimension named `coeff` - that is not good but would be a good name? :face_in_clouds:

### DIFF
--- a/mesmer/stats/_power_transformer.py
+++ b/mesmer/stats/_power_transformer.py
@@ -3,7 +3,7 @@ import xarray as xr
 from scipy.optimize import minimize
 
 
-def lambda_function(xi_0, xi_1, local_yearly_T):
+def lambda_function(coeffs, local_yearly_T):
     r"""Use logistic function to calculate lambda depending on the local yearly
     values. The function is defined as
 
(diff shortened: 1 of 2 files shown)


veni-vidi-vici-dormivi:
Thinking about the fact that we might want to merge all parameters into one dataset in the end we could go for something like pt_parameters.

mathause:
`params` sounds good or maybe `lambda_coeffs`. I thought about naming the dimension `beta` as regression params are often called $\beta$  (breaks down a bit for the constant lambda)

veni-vidi-vici-dormivi:
Hm yeah `lambda_coeffs` is nice. Why not `coeff` as dim? That's also what we have for the harmonic model.

mathause:
> Hm yeah `lambda_coeffs` is nice. Why not `coeff` as dim? That's also what we have for the harmonic model.

Yup - let's go for this - thanks.

veni-vidi-vici-dormivi (approved):
Cool thanks!

veni-vidi-vici-dormivi:
Thanks!

PR #521, opened 2024-09-10
Add Example for several Scenarios

veni-vidi-vici-dormivi:
if this works could also add an integration test for this
 - [ ] Closes add later
 - [ ] Tests added


veni-vidi-vici-dormivi:
image

Okay so when I plot the residuals after removing the global trend (including volcanic forcing) where I treat the historical members as their own scenario I get this "mismatch" at the transition from historical to projected period. At the moment I would say that this is not pretty but actually, for the fitting it should be fine as long as we keep treating the historical data as its own scenario for the AR processes. For the linear regressions and the variances/covariances it doesn't matter since these do not consider time dependency. But I would be happy if a second brain went over this as well @mathause 🙂 

We should point out however that in the emulation process one should use a continuous time series not ensure continuity of the realization.

veni-vidi-vici-dormivi:
Okay, I'm actually surprisingly happy with this approach and impressed by what xarray and datatree can do. I feel like the data tree approach I went for here (holding one dataset per scenario that holds the members along the dimensions) is nice.

Nevertheless, I want to rewrite the autoregression functions to work on data trees instead of the arg list. For the linear regression and covariance we could think about implementing functions that take care of the stacking and weighting automatically. Actually I think this would be quite fun.

But I want to focus on MESMER-X for the rest of the week.

veni-vidi-vici-dormivi:
@yquilcaille You can use this now. All functionality should stay the same as it is in here now, just that some of the manual data prepping I do will be moved into functions, which needs more time to implement cleanly.

One thing, if you calibrate on all the ESMs, could you tell me if you ever run into singular correlation matrices when fitting for the best localization radius? At the moment this should abort the fitting and we are still debating if it is worth to implement a version where we singular matrices are allowed. Thank you!

yquilcaille (approved):
Thanks @veni-vidi-vici-dormivi! The "surfer" looks good, no problem to add it. I agree that the preparation of the data should be moved into functions with the future cleaning. Also, some users may benefit from easy wrappers, like one for training, one for emulation.

I will now use this surfer to prepare the training of all ESMs and emulations for FASTMIP. Promised, if any issue appears on the singular matrices, I will let you know :)

mathause:
Thanks! Cool that this works & sorry for the late reply. I would like to see some changes before merging, though,

- Should we use the example datasets so it's self-contained? Maybe could mention how to use with cmip6-ng.
- Did you ever double check this is consistent with Lea's results?
- rename notebook to e.g. `example_mesmer_multi_ens_multi_scen.ipynb`?
- There is quite a bit of clean-up possible
  - unused functionality and imports (but see the suggestions above)
  - in Cell 12 you manipulate the data in a function call which is not nice (in a tutorial)
  - there is more but commenting a notebook is annoying - I might go over the notebook later, but would appreciate if you gave it a first pass
- the emulations have the same variability. Haven't we discussed this?
  - can you gather the emulation part into a function (in the notebook) - maybe needs a function for the variability and one for the trend?
  - should the historical part have for two different scenarios have the same variability? I think both should be possible

mathause:
TODO: check the status of datatree in xarray - it would be good if we can enable using it from xarray (`DataTree` and `map_over_subtree` is now available from the main xarray namespace) but needs a relatively new version.

veni-vidi-vici-dormivi:
> Should we use the example datasets so it's self-contained? Maybe could mention how to use with cmip6-ng.

Yes absolutely, have done this locally already, will push it soon. I am currently working on implementing the data tree approach in the repo and moving this into the integration tests.

> Did you ever double check this is consistent with Lea's results?

It is not because I treat the historical period as a completely independent scenario, i.e. I smooth historical and scenario separately thus leading to different values around the transition from historical to future period. This leads to different values than Lea's in the smoothed global mean and thus the residuals and everything thereafter, thus all the parameters. What do you think about this? I think that it is more elegant as there is no duplication of the historical period. As I see it, Lea solved this by taking the median over the scenario hists before: 

https://github.com/MESMER-group/mesmer/blob/72d2fd91a36c2a6dcd773174dc0f2c05af5d0a76/mesmer/calibrate_mesmer/train_gt.py#L100-L105

> rename notebook to e.g. example_mesmer_multi_ens_multi_scen.ipynb?

Agree. Will do.

> There is quite a bit of clean-up possible

Will get back to this later

> the emulations have the same variability. Haven't we discussed this?
can you gather the emulation part into a function (in the notebook) - maybe needs a function for the variability and one for the trend?
should the historical part have for two different scenarios have the same variability? I think both should be possible

Ah right. If we want different historical variability we just need different seeds for each scenario right, so both is possible depending on the seed? 
Yes, good idea to gather it into a function in the notebook but not in the repo, I like that. I need to think about how to write it so that we can potentially reuse the trend.

veni-vidi-vici-dormivi:
What I did now is a bit wild but I think it could pay off:

I wanted to add the example to the readthedocs documentation. `nbspinx` can render notebooks in the documentation but it is apparently not possible that the notebook lives outside the doc source directory (I really tried hard). Thus I decided to create a `tutorial` folder inside the docs and put the example there. This works nicely and I think it is very valuable to show the tutorial on the documentation page. However, this kind of hides away the notebook from potential developers/users that want to run it on their own. So to combat this I decided to move all the documentation sources one level up from `docs/source` directly into `docs` I think this is okay since there was anyways only the source folder and the makefile in there to start with. Please @mathause let me know what you think about this.

veni-vidi-vici-dormivi:
> Could a softlink work?

No for some reason it marks a softlink as an unreadable document.

You mean we could set up a workflow that copies and builds the notebook on every push? I don't build it locally, I use nbsphinx to build it on make html.

mathause:
> > Could a softlink work?
> 
> No @mathause for some reason it marks a softlink as an unreadable document.

Pity :unamused: 

> You mean we could set up a workflow that copies and builds the notebook on every push? I don't build it locally, I use nbsphinx to build it on make html.

Ah right (I thought you push the rendered notebook because it still has the output. For regionmask I use myst_nb to do the same because I make use of some directives (admonitions, e.g. `warning` that don't exist in markdown. No TODO, just FYI.). I fear that it will be slow, e.g. if we want to add a notebook for the more complex modules, like MESMER-M, but let's not worry about this now.

But you were able to keep the other files in `source`?

veni-vidi-vici-dormivi:
> Pity

Yes. It was very frustrating to find out.

> because it still has the output. 

Yes because I didn`t clear it in the Notebook. Interesting that you use myst_nb, I just used the first recommendation, it was not obvious to me which one is better.

> But you were able to keep the other files in source

No I moved everything one level up and deleted the source folder. I see how the docs folder might be a bit cluttered now but I like it better actually because the documentation is more accessible this way.

mathause:
> > Pity
> 
> Yes. It was very frustrating to find out.

For an md file it works :man_shrugging: but maybe not for nbsphinx? Could also try with myst_nb...

https://git.iac.ethz.ch/ip_python/ip_python/-/tree/main/docs/source/climate_risk?ref_type=heads


> Interesting that you use myst_nb, I just used the first recommendation, it was not obvious to me which one is better.

It's not obvious. I used myst because it offers more formatting options than pure markdown. See [example in regionmask docs](https://regionmask.readthedocs.io/en/stable/notebooks/mask_2D.html): admonitions (the Note, and Tip etc. sections) and substitutions (the topmost `Note` section which is the same in all notebooks). I don't particularly like the syntax, though. Also: regionmask/regionmask/issues/532

Also I had problems with nbconvert - it requires pandoc and that cannot be installed via pip and the version ubuntu offers is rather old. Not sure why this problem does not surface here.

mathause:
> Also I had problems with nbconvert - it requires pandoc and that cannot be installed via pip and the version ubuntu offers is rather old. Not sure why this problem does not surface here.

Now I understand what's happening - you "only" convert the executed notebook to rst. While I execute and convert it (on RTD). But that is probably too slow here.

mathause:
I just adapted the notebook for `xr.DataTree`. It looks great, this will be an important addition to our examples and docs. #632 will allow to remove many `map_over_subtree` calls for a small clean up. I need to go over the contents again (not only  making it work).

I am still considering if this should be in the docs or in examples.

mathause:
Ok I reverted the changes to include the tutorial on RTD. It was mostly

```bash
git mv docs/tutorials/tutorial_mesmer_multiple_scenarios.ipynb examples/
git reset main -- docs
```

and then a lot of cleanup...

veni-vidi-vici-dormivi:
Ok this is ready now imo @mathause

mathause (approved):

mathause:
thanks a lot @veni-vidi-vici-dormivi

veni-vidi-vici-dormivi:
ahh omg 🥰 thank you! Do we now want to go for putting it into the docs? Do you already have a blueprint / plan for that?

PR #523, opened 2024-09-13
Update conda environment creation

### LINKED ISSUE #315: getting started with MESMER
Some instructions to create the necessary environment for MESMER. Theses steps have to be done only once.

Step 0 is specific to the setup at ETH/ IAC:

0. Make sure you have conda/ mamba available
   - Add the line `module load conda` to the `~/.bashrc` file
   - Open a Terminal (`File -> New -> Terminal`)
   - Call `mamba init`
   - Close the Terminal and open a new one
   - The terminal should now say `(base) username@host:~>`

---

1. Obtain the MESMER source code by cloning it
   - Run `git clone https://github.com/MESMER-group/mesmer`
   - You should now have a `mesmer` folder
2. Create a mamba/ conda environment to work with MESMER
   - Enter the mesmer folder `cd mesmer`
   - Run `mamba env create -n mesmer_dev -f environment.yml`
   - This takes some minutes and creates a new environment with the name `mesmer_dev`
3. Install additional dependencies:
   - Activate the new environment `mamba activate mesmer_dev`
   - Make sure the command line now says: `(mesmer_dev) username@host:~/mesmer>`
   - Install ipykernel so you can use it from JupyterHub `mamba install ipykernel matplotlib`
   - Install filefinder: `python -m pip install git+https://github.com/mathause/filefinder/`
4. Install mesmer itself in development mode (you need to be in the mesmer folder):
   - `python -m pip install -e .`
5. To check that mesmer was correctly installed you can run all test:
   - `pytest . --all`

veni-vidi-vici-dormivi:
I didn't test if the instructions to install the environment actually work, so here are the fixes

 - [x] Closes #315 

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -58,6 +58,7 @@ Bug fixes
 
 Documentation
 ^^^^^^^^^^^^^
+- Updated and extended the development Guide (`#511 <https://github.com/MESMER-group/mesmer/pull/511>`_, `#523 <https://github.com/MESMER-group/mesmer/pull/523>`_)
 
 Internal Changes
 ^^^^^^^^^^^^^^^^
--- a/docs/source/development.rst
+++ b/docs/source/development.rst
@@ -59,16 +59,16 @@ Follow the steps below (if any of these tools are unfamiliar, please see the res
 
    a. Change your current directory to MESMER's root directory (i.e., the one that contains ``README.rst``), ``cd mesmer``.
 
-   b. Create a conda environment to use with MESMER: ``conda create --name mesmer``.
+   b. Create a conda environment to use with MESMER and install all the dependencies: ``conda env create -n mesmer_dev -f environment.yml``.
 
-   c. Activate your conda environment: ``conda activate mesmer``. This is important for the next step, otherwise the packages will be installed in your base environment.
+   c. Activate your conda environment: ``conda activate mesmer``. This is important for the next step, otherwise the packages will be installed in your base environment. Make sure the command line now says: ``(mesmer_dev) username@host:~/mesmer>``.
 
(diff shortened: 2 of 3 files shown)


veni-vidi-vici-dormivi:
Also the installation.rst states the dependencies but these are missing:
- cartopy=0.22
 - joblib=1.3
 - netcdf4=1.6
 - packaging=23.1
 - properscoring=0.1
 - shapely=2.0

Do we want to add them?

veni-vidi-vici-dormivi:
> I am a bit unsure about this ATM. I would probably go the other way round and have the installations instructions in the installation.md and point to that from the devel docs.

Hm but the installation instructions for the normal user are here, only. the ones for developers are in the development guide... I think I like this a little better because then developers need only one page for everything...?

> Where are they missing from? I see them in environment.yml

They are missing from the list in the installation.md. I can't link it from the PR because I didn't change anything there, but it's here: 

https://github.com/MESMER-group/mesmer/blob/6c800e5c3b54f8457ffbcd5ec8d7b59411e74f3b/docs/source/installation.rst?plain=1#L7-L17

> Then there are some deps that are required for using mesmer but not for the test runs (matplotlob, ipython) - although it adds redundancies we should maybe add everything for development in environment.yml and have the reduced set in ci/requirements.env.yml.

Interesting point. Atm they are given in the developer guide as add ons, for me this is fine.

mathause (approved):
Ok fair enough. Currently I would recommend a clone and editable install for most users, so I thought we could add this to the general instructions. But ok to leave as is for now.

>> Where are they missing from? I see them in environment.yml

> They are missing from the list in the installation.md

Ah thanks. Can you add

- netcdf4
- joblib
- properscoring

to the deps. And add a new section with optional deps:

- cartopy
- matplotlib
- nc-time-axis

PR #530, opened 2024-09-30
mesmer-x integration test: mark both as slow

mathause:
I am regularly running `pytest .` and want to be able to keep doing that. A runtime of ~2 min is too long for that, so I have to mark this test as slow as well. Optimally would speed this up considerably, but that is for another time ;-)

This test still runs on GHA


mathause:
Running the full test suite now takes 7 min...

mathause:
thanks

PR #526, opened 2024-09-18
mesmer_x: add tests for `Expression`

mathause:
- [ ] Closes #xxx
 - [ ] Tests added
 - [ ] Fully documented, including `CHANGELOG.rst`

Some first unit tests for `Expression`. Exposed #525. What I am not quite sure yet: should I also do some refactoring here?

PR #531, opened 2024-09-30
`Expression`: disallow bare functions

mathause:
As correctly guessed by @veni-vidi-vici-dormivi in https://github.com/MESMER-group/mesmer/pull/526#pullrequestreview-2317904070 I want to disallow passing bare functions to an expression, e.g. `loc=mean()`. This function would then be translated to `loc=np.mean()`.

@yquilcaille would like to keep this functionality (https://github.com/MESMER-group/mesmer/pull/526#issuecomment-2363569175).

I still want to remove it because it's **not difficult to fix** for the user, but difficult to explain and not obvious what happens. Removing also makes the code easier to understand.

### DIFF
--- a/mesmer/mesmer_x/train_utils_mesmerx.py
+++ b/mesmer/mesmer_x/train_utils_mesmerx.py
@@ -247,7 +247,7 @@ class Expression:
                 expr = expr.replace(x, "")
 
             # reading this condensed expression to find terms to replace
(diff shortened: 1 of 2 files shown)


yquilcaille (approved):
I don't mind not allowing bare functions, there are certainly good reasons for that ^^
Looks good to me!

PR #534, opened 2024-10-01
`Expression`: test `evaluate` method

mathause:
Added tests for `expr.evaluate` - I will do some more refactoring of the method later but wanted to get some tests in first.


mathause:
- I don't test `forced_shape` - not sure if that is really needed - do we ever pass arrays with more than 1D?
- I am also not sure that I test every possibility
- do we actually need to broadcast scalar coeffs? I expect scalars to just work. Would be good to have a failure case for that.

veni-vidi-vici-dormivi (approved):

mathause:
Thanks

PR #535, opened 2024-10-01
`Expression.evaluate`: avoid `exec` by using `locals`

mathause:
We can pass `locals` ("locally defined variables") as argument to `eval` - this avoids `exec` and "magically defined variables". I think this helps a lot with the readability of the code.

- `locals` is a positional-only argument, so the `None` is required
- as `coefficients_values` or `inputs_values` can be a `Dataset` we cannot use dict union (`d1 | d2`) and need to splat them
- I suggest to merge #534 first so to have the tests first (therefore only draft PR at the moment)

### DIFF
--- a/mesmer/mesmer_x/train_utils_mesmerx.py
+++ b/mesmer/mesmer_x/train_utils_mesmerx.py
@@ -353,19 +353,14 @@ class Expression:
         if len(shapes) > 1:
             raise ValueError("Different shapes of inputs detected.")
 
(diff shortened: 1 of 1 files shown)


mathause:
Given @yquilcaille is super busy for another two weeks, maybe @veni-vidi-vici-dormivi can review?

veni-vidi-vici-dormivi (approved):

mathause:
Thanks!

PR #536, opened 2024-10-01
mesmer-x: remove old train and draw modules

mathause:
We have `TODO: delete` in both files, so I suggest to do this - this will remove some mental load. If there is a reason to keep them - no worries I can close this PR. Otherwise I suggest to go ahead - we can always get it back.

Unrelated question: how would we draw emulations with the new mesmer-x code base?


veni-vidi-vici-dormivi:
Not sure if we still need the back transform in create emus. Couldn't find it anywhere else?

mathause:
Is it 

https://github.com/MESMER-group/mesmer/blob/72d2fd91a36c2a6dcd773174dc0f2c05af5d0a76/mesmer/mesmer_x/train_utils_mesmerx.py#L407

?

veni-vidi-vici-dormivi:
Yeah I don't fully understand, this might just be any distribution -> normal distribution but not back? Ah but maybe when I just switch start and end expression it works?

mathause:
Gentle ping @yquilcaille - do you agree to remove these files?

mathause:
Ok merging. We can always revert.

PR #537, opened 2024-10-02
Add calibration integration test for multiple scenarios

### LINKED ISSUE #18: Change to saving bundle using more standard format
At the moment we use `.pkl` files which are a bit dangerous

### LINKED ISSUE #407: Extend calibration integration tests
The `test_calibrate_mesmer_newcodepath` (see https://github.com/MESMER-group/mesmer/pull/400) now only covers the very basic test case of one ensemble member, one scenario.
For testing the other case we first need to make progress on the internal data structure, see https://github.com/MESMER-group/mesmer/issues/106.

veni-vidi-vici-dormivi:
Add calibration integration tests for several scenarios using `DataTree` as a data structure. Towards #106. 
BREAKING CHANGE: we now treat historical runs as a completely separate scenario throughout the whole calibration. The difference to Lea's version is that we do not apply the loess smoother to a continuous hist-future timeseries but to the historical runs and the future runs individually. The rest is the same (for the autoregression Lea also treated them as separate scenarios, and for the Linear Regression it does not make a difference).

 - [x] Closes #407 
 - [x]  Closes #18
 - [x] Tests added

Note that the comments at the beginning of the PR are remainders from discussions about implementing `DataTree` as a data structure when this was still a draft PR.


mathause:
Unfortunately filefinder is not on conda - it's not even on pypi. The problem is that there is already a package with the same name on pypi - so I wasn't sure what to do. We can either work on getting it published (under another name) - or install it from github.

veni-vidi-vici-dormivi:
Hm I see... What about `feistyfilefinder`? That's the best thing I was able to come up with... Installing it from GitHub is also fine I guess, need to check how that works

mathause:
https://github.com/C2SM/pyvis/blob/f7e509b9627d36cb09d4eee9ed1f3a84c6eeadda/environment.yml#L42

(using `*.tar.gz` (or `*.zip`) has the advantage that git does not need to be installed but we have to manually update the versions, while `git+https://...` installs the dev version but requires git on the `PATH`)

veni-vidi-vici-dormivi:
I think having to manually update it is fine, since we would be the ones updating filefinder.

mathause:
Yes, that's also my preferred option (could have mentioned that above...)

mathause:
- how much work would it be to continue allowing lists to the AR functions? Along the lines of
   ```diff
   -def _select_ar_order_scen_ens(*obj, ...)
   +def _select_ar_order_scen_ens(obj, ...)
   +    if isintance(obj, DataTree):
   +        ...
   ```
- `xr.DataTree` is now available, so I feel it should now be usable for the newest carray release - can you check the xarray changelog if you find something on that? (best would be the version when it's available)
- if it is usable it would be cool to prefer it over datatree (if using a new-enough version) - but let's discuss first

- p.s. it needs a `closes` before each item in the list otherwise it does not work (only one item is listed under "Development")

veni-vidi-vici-dormivi:
> xr.DataTree is now available, so I feel it should now be usable for newer xarray releases - can you check the xarray changelog if you find something on that? (best would be the version when it's available)

Yes indeed, it is available since v2024.05. We are currently requiring at least v2023.07. Should I try with v2024.05?

veni-vidi-vici-dormivi:
> p.s. it needs a closes before each item in the list otherwise it does not work (only one item is listed under "Development")

Sorry, what is this in reference to?

mathause:
> > xr.DataTree is now available, so I feel it should now be usable for newer xarray releases - can you check the xarray changelog if you find something on that? (best would be the version when it's available)
> 
> Yes indeed, it is available since v2024.05. We are currently requiring at least v2023.07. Should I try with v2024.05?

I first thought we could support both - but on second thought I don't think it's worth it, so yes.

- we should probably add a nice error message though (`try: import ...`)
- will have to update the min-deps-check so it does not complain (probably needs to be `"xarray": (2024, 5),`)
   https://github.com/MESMER-group/mesmer/blob/72d2fd91a36c2a6dcd773174dc0f2c05af5d0a76/ci/min_deps_check.py#L37


> > p.s. it needs a closes before each item in the list otherwise it does not work (only one item is listed under "Development")
> 
> Sorry, what is this in reference to?

The list of issues that are closed by this PR - see the side bar

image

mathause:
and by the way: awesome :+1:

veni-vidi-vici-dormivi:
Ah the `DataTree` will only be moved to the main namespace in the newest release:

https://github.com/pydata/xarray/commit/18e5c87696c2233202bc67066408e6f495d0327d#diff-66d415f4d4a5d0969b40e35b86cbf67612bc3b88c7f02957a550f12df7e0e14eR19-R24

Before that we would need to import it from `xarray.core.datatree` and `map_over_subtree` from `xarray.core.datatree_mapping` 🙈

mathause:
As discussed - in this case it does not make sense to require `DataTree` from xarray - so my suggestion is we merge this PR with datatree-xarray and add a compatibility layer in another PR.

veni-vidi-vici-dormivi:
I don't know what do to anymore what to do about the issue with windows... A way out could be to activate the conda environment and only installing filefinder afterwards but then we would have to put this in the ci-workflow and I don't like that...

mathause:
> I don't know what do to anymore what to do about the issue with windows... A way out could be to activate the conda environment and only installing filefinder afterwards but then we would have to put this in the ci-workflow and I don't like that...

We use bash as shell on all 3 OS - it might be possible that we can get it to work on another shell on windows. But that might bring other problems with it. Anyways as discussed I will test that locally on my windows laptop and see if I can reproduce the problem.

mesmer/stats/_linear_regression.py
```diff
@@ -319,3 +356,92 @@ def _fit_linear_regression_np(predictors, target, weights=None, fit_intercept=Tr
         intercepts = np.zeros_like(coefficients[:, :1])
 
     return np.hstack([intercepts, coefficients])
+
+
+def prep_linear_regression_data(
```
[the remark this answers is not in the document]
mathause: or maybe `stack_...`? to convey a bit better what is actually done?

mesmer/stats/_linear_regression.py
```diff
@@ -319,3 +356,92 @@ def _fit_linear_regression_np(predictors, target, weights=None, fit_intercept=Tr
...
+
+    # exclude target dimensions from broadcasting which are not in the stacking_dims
+    exclude_dim = set(target.leaves[0].ds.dims) - set(stacking_dims)
+
+    # predictors need to be
+    for key, subtree in predictors.items():
+        # 1) broadcast to target
+        broadcasted = subtree.broadcast_like(target, exclude=exclude_dim)
+        # 2) collapsed into DataSets
+        ds = collapse_datatree_into_dataset(broadcasted, dim=collapse_dim)
```
mathause, line 421:
```suggestion
        broadcast = subtree.broadcast_like(target, exclude=exclude_dim)
        # 2) collapsed into DataSets
        ds = collapse_datatree_into_dataset(broadcast, dim=collapse_dim)
```

I think `broadcasted` is not correct

mesmer/stats/_linear_regression.py
```diff
@@ -319,3 +356,92 @@ def _fit_linear_regression_np(predictors, target, weights=None, fit_intercept=Tr
...
+            )
+        )
+
+    # target needs to be
+    # 1) collapsed into DataSet
+    target_ds = collapse_datatree_into_dataset(target, dim=collapse_dim)
+    # 2) stacked
+    target_stacked = target_ds.stack(
+        {stacked_dim: stacking_dims}, create_index=False
+    ).dropna(dim=stacked_dim)
```
mathause, line 435: can you define `stack_dim` at the beginning of the function and do this in two steps - could be a easier to read

```python
stack_dim = {stacked_dim: stacking_dims}

...

target_stacked = target_ds.stack(stack_dim, create_index=False)
target_stacked = target_stacked.dropna(dim=stacked_dim)
```

mesmer/stats/_linear_regression.py
```diff
@@ -319,3 +356,92 @@ def _fit_linear_regression_np(predictors, target, weights=None, fit_intercept=Tr
...
+
+    # weights need to be
+    if weights is not None:
+        # 1) collapsed into DataSet
+        weights_ds = collapse_datatree_into_dataset(weights, dim=collapse_dim)
+        # 2) stacked
+        weights_stacked = weights_ds.stack(
+            {stacked_dim: stacking_dims}, create_index=False
+        ).dropna(dim=stacked_dim)
+        return predictors, target_stacked, weights_stacked
```
mathause, line 445:
```suggestion

        return predictors, target_stacked, weights_stacked
```

mesmer/stats/_linear_regression.py
```diff
@@ -319,3 +356,92 @@ def _fit_linear_regression_np(predictors, target, weights=None, fit_intercept=Tr
...
+
+    return predictors, target_stacked
```
mathause, line 447: functions with variable number of return values are usually impractical to use - can you return `None` as the weights? Or maybe a dict that can be unpacked? Or a separate fcn for the weights?

mathause (commented):
I had a first look and added some comments. It's too big to really understand everything but think it's going in the right direction.

mesmer/core/weighted.py
```diff
@@ -106,3 +108,77 @@ def global_mean(data, weights=None, x_dim="lon", y_dim="lat"):
...
+
+        data_vars = list(ds.keys())
+        if len(data_vars) > 1:
+            raise ValueError("Dataset must have only one data variable.")
+
+        # Get the dimensions to calculate the weights for and make sure they are in the right order
+        dims = [
+            dim
+            for dim in ds[data_vars[0]].dims
+            if exclude is None or dim not in exclude
```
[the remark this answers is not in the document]
mathause:
```python
if exclude is None:
    exclude = set()
```
maybe even clearer

mesmer/core/volc.py
```diff
@@ -98,7 +99,7 @@ def fit_volcanic_influence(tas_residuals, hist_period, *, dim="time", version="2
 
     # TODO: name of 'aod'
     lr.fit(
-        predictors={"aod": aod},
+        predictors=DataTree.from_dict({"aod": aod}),
```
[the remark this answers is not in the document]
veni-vidi-vici-dormivi: You are right, it's really not. But only because we cannot save `xr.DataArrays` in `DataTree`s but only `xr.Datasets`. I changed it in the spirit of using `DataTree`s as datastructure.

mesmer/core/utils.py
```diff
@@ -240,3 +242,88 @@ def _check_dataarray_form(
...
+    # Concatenate datasets along the specified dimension
+    ds = xr.concat(datasets, dim=dim)
+    ds = ds.assign_coords(
+        {dim: [subtree.name for subtree in dt.subtree if not subtree.is_empty]}
+    )
+
+    return ds
+
+
+def _datatree_to_arraydict(dt: DataTree) -> dict[str, xr.DataArray]:
```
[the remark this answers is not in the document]
veni-vidi-vici-dormivi: Yes.

mesmer/stats/_linear_regression.py
veni-vidi-vici-dormivi: 😅

mesmer/stats/_linear_regression.py
veni-vidi-vici-dormivi: I went for returning None now.

mesmer/__init__.py
```diff
@@ -9,16 +9,17 @@
...
-from .core import geospatial, grid, mask, volc, weighted
+from .core import geospatial, grid, mask, utils, volc, weighted
 
 # "legacy" modules
 __all__ = [
     "calibrate_mesmer",
     "create_emulations",
     "io",
-    "utils",
+    "old_utils",
```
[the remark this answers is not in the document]
veni-vidi-vici-dormivi: Yeah actually a DataTree utils would be nice. I renamed it because I got annoyed by how complicated it was to. import things from there but I agree that this should be cleaner

mathause:
A new xarray version that contains `DataTree` was released (just FYI)

veni-vidi-vici-dormivi:
Okay I would like to take this apart into individual PRs, mainly:
- add DataTree dependency
- warn on superfluous predictors
- (typing fixes)
- datatree utils
- autoregression
- weighting function
- linear regression
- calibration integration tests
- example
- emulation integration tests

My preferred approach would be to just copy paste the files into new PRs and not make the hassle of cherry picking everything. Are you okay with that @mathause?

mathause:
Yes that's good for me. You could do partial commits (but would need to `git reset main` (I think) to get everything in un-commited form again). I am not sure it's worth it but if you want to try it.


https://www.youtube.com/watch?v=UJ5fpaeZWsI

tests/integration/test_calibrate_mesmer_newcodepath.py
```diff
@@ -86,46 +89,104 @@ def test_calibrate_mesmer(
...
+        scenario="historical",
+        model=esm,
+        resolution="g025",
+        time_res="ann",
+        member=unique_scen_members,
+    )
 
-    fN_hist = path_tas / f"tas_ann_{esm}_historical_r1i1p1f1_g025.nc"
-    fN_proj = path_tas / f"tas_ann_{esm}_{scenario}_r1i1p1f1_g025.nc"
+    fc_all = FileContainer(pandas.concat([fc_hist.df, fc_scens.df]))
```
mathause, line 127: Do we need to release filefisher first?

tests/integration/test_calibrate_mesmer_newcodepath.py
```diff
@@ -86,46 +89,104 @@ def test_calibrate_mesmer(
...
     TEST_DATA_PATH = pathlib.Path(test_data_root_dir)
-    TEST_PATH = TEST_DATA_PATH / "output" / "tas" / "one_scen_one_ens"
+    TEST_PATH = TEST_DATA_PATH / "output" / outname
 
     cmip_data_path = (
         TEST_DATA_PATH / "calibrate-coarse-grid" / f"cmip{test_cmip_generation}-ng"
     )
 
-    path_tas = cmip_data_path / "tas" / "ann" / "g025"
+    CMIP_FILEFINDER = FileFinder(
```
mathause, line 106: we should extract loading cmip6-ng to a helper function (but let's keep for now)

tests/integration/test_calibrate_mesmer_newcodepath.py
```diff
@@ -86,46 +89,104 @@ def test_calibrate_mesmer(
...
+        resolution="g025",
+        time_res="ann",
+        member=unique_scen_members,
+    )
 
-    fN_hist = path_tas / f"tas_ann_{esm}_historical_r1i1p1f1_g025.nc"
-    fN_proj = path_tas / f"tas_ann_{esm}_{scenario}_r1i1p1f1_g025.nc"
+    fc_all = FileContainer(pandas.concat([fc_hist.df, fc_scens.df]))
+
+    scenarios_whist = scenarios.copy()
```
mathause, line 129:
```suggestion
    scenarios_incl_hist = scenarios.copy()
```
I think that's easier to understand

tests/integration/test_calibrate_mesmer_newcodepath.py
```diff
@@ -86,46 +89,104 @@ def test_calibrate_mesmer(
...
+    TEST_PATH = TEST_DATA_PATH / "output" / outname
 
     cmip_data_path = (
         TEST_DATA_PATH / "calibrate-coarse-grid" / f"cmip{test_cmip_generation}-ng"
     )
 
-    path_tas = cmip_data_path / "tas" / "ann" / "g025"
+    CMIP_FILEFINDER = FileFinder(
+        path_pattern=cmip_data_path / "{variable}/{time_res}/{resolution}",  # type: ignore
+        file_pattern="{variable}_{time_res}_{model}_{scenario}_{member}_{resolution}.nc",
```
mathause, line 108: Rename `member` to `ens`, `ensmember` or `variant` (which is the official cmip name)?

tests/integration/test_calibrate_mesmer_newcodepath.py
```diff
@@ -86,46 +89,104 @@ def test_calibrate_mesmer(
...
+            # drop unnecessary variables
+            ds = ds.drop_vars(["height", "time_bnds", "file_qf"], errors="ignore")
+            # assign member-ID as coordinate
+            ds = ds.assign_coords({"member": meta["member"]})
+            members.append(ds)
+
+        # create a Dataset that holds each member along the member dimension
+        scen_data = xr.concat(members, dim="member")
+        # put the scenario dataset into the DataTree
+        dt[f"{scen}"] = DataTree(scen_data)
```
mathause, line 150: f-string not really necessary

tests/integration/test_calibrate_mesmer_newcodepath.py
```diff
@@ -134,95 +195,118 @@ def mask_and_stack(ds, threshold_land):
...
-    tas_stacked = mask_and_stack(tas, threshold_land=THRESHOLD_LAND)
+    tas_stacked = map_over_subtree(mask_and_stack)(
+        tas_anoms, threshold_land=THRESHOLD_LAND
+    )
 
     # train global trend module
-    tas_globmean_lowess = mesmer.stats.lowess(
-        tas_globmean, "time", n_steps=50, use_coords=False
+    tas_globmean_smoothed = map_over_subtree(mesmer.stats.lowess)(
+        tas_globmean.mean(dim="member"), "time", n_steps=50, use_coords=False
```
mathause, line 204: can you put `tas_globmean_ensmean = tas_globmean.mean(dim="member")` on a separate line

tests/integration/test_calibrate_mesmer_newcodepath.py
```diff
@@ -134,95 +195,118 @@ def mask_and_stack(ds, threshold_land):
...
-        *data, dim="time", ens_dim="ens", lags=ar_order
+        tas_resid_novolc, dim="time", ens_dim="member", lags=ar_order
     )
 
+    if use_hfds:
+        hfds_ref = dt_hfds["historical"].sel(time=REFERENCE_PERIOD).mean("time")
+        hfds_anoms = dt_hfds - hfds_ref.ds
+        hfds_globmean = map_over_subtree(mesmer.weighted.global_mean)(hfds_anoms)
+        hfds_globmean_smoothed = map_over_subtree(mesmer.stats.lowess)(
+            hfds_globmean.mean(dim="member"), "time", n_steps=50, use_coords=False
```
mathause, line 236: and this as well

mathause (commented):
Looks good.

veni-vidi-vici-dormivi:
Okay I tested how this change in the method/the parameters changes the emulations: 
Below is a QQ plot for emulations created with the old parameters and the old code path ("old") and emulations with the new params and new codepath for the configuration ("new"). The parameters are from training on the two coarse members of ssp585 (where there is more difference in the parameters than for ssp126) and taking a dummy forcing for emulation (linear increase from 0 to 1 during historical and from 1 to 2 during future period). This shows 1000 emulations with their time and gridcell dimensions stacked. I used the same seeds for old and new.

image

I think it looks good, although there might be some slight underestimation of cool extremes by the new emulations in comparison to the old ones I guess? (I'm new with QQ plots) 

This is what two timeseries look like at gridcell 30 for one realization:
image

I would say it looks fine 🤗

tests/integration/test_calibrate_mesmer_newcodepath.py
veni-vidi-vici-dormivi: image

apparently variant

veni-vidi-vici-dormivi: or variant label if we want to be fully consistent

veni-vidi-vici-dormivi: but I would want to go for variant

mathause: I know - sorry, my sentence was not clear... So yes that would be good for consistency - thing is I don't particularly like `variant` :expressionless: (This string also encodes different physics particularization or forcing, and not only a different realization (and with only the third it would strictly be a ensemble member), however, I never cared about this and pooled them all, while you could argue that a different physics variant is a different model while a different forcing variant is strictly speaking a different scenario... I never considered this in my analyses and they are not often used)

Not sure this made sense...

mathause: See footnote 8 in https://docs.google.com/document/d/1h0r8RZr_f3-8egBMMh7aqLwy3snpD6_MrDz1q8n5XUk/edit?usp=sharing

veni-vidi-vici-dormivi: ah yes I saw the foot note. I see, so I guess `variant` is fine but for consistency we should then rename it everywhere and I also dont like it too much. I would stick with naming it variant in the FileFinder and member everywhere else.

mathause (approved):
I was also very happy with this PR when I reviewed it - so the only open question is the naming of the ensemble members

- `variant`
- `member`
- `ensemble_member`
- `ens`
- `ensmember`

My personal preference would be `ensmember` but I get how that is very personal, so let's go with your choice of `member`. I would probably also keep this consistent with filefisher and not use `variant` there.

veni-vidi-vici-dormivi:
Okay last question: do you agree with the file naming structure 
```
PARAM_FILEFINDER = FileFinder(
        path_pattern=TEST_PATH / "test-params/{scope}/{param_type}",
        file_pattern="params_{short_type}_{method}_{esm}_{scen}.nc",
    )
```
So that for example the file would be called `/mesmer/tests/test-data/output/tas/multi_scen_multi_ens/test-params/global/global_trend/params_gt_LOWESS_OLSVOLC_saod_IPSL-CM6A-LR_ssp126-ssp585.nc`
I would personally like to remove short_type and method

mathause:
I haven't thought much about the naming of the output - good that you do :slightly_smiling_face: 

> I would personally like to remove short_type and method and replace method with the module maybe like volcanic.

Can you type that out for an example?

mathause:
So `module` is the mesmer module and not the python module. And yes I am also for getting rid of the short type - they are not very intuitive for me.

* Maybe `covariance` -> `cov`? or `localized-covariance` -> `covariance`?
* it's good to use `_` for the categories and `-` within the categories

veni-vidi-vici-dormivi:
I also looked at your suggestion in https://github.com/MESMER-group/mesmer/issues/65#issuecomment-959123361, we could also omit the module folder. What do you think @mathause. I agree with only using `covariance`.

veni-vidi-vici-dormivi:
Okay I'm merging now, we can have a discussion about the naming later at some point, as I really want to get some more important things done now.

veni-vidi-vici-dormivi:
I came across something I did here @mathause : I average the global mean over all ensemble members in the beginning. Consequences:
1. the volcanic parameters are trained on residuals of each member to the smoothed mean over all members 
2. so is the global ar process
3. I use the smoothed global mean averaged over all members as predictor for each member in the linear regression. 

Question: is this what Lea did? I looked into the legacy code and feel like she actually uses each members global smoothed mean as predictor for its corresponding gridded member. Should we implement that?

mathause:
I know that she averages over all ensemble members for the loess smoothing:

https://github.com/MESMER-group/mesmer/blob/b2f2ebd052b50cfce61152a1780f4ec06668091e/mesmer/calibrate_mesmer/train_gt.py#L172

but I don't think she does that for the other stuff. I had a look and got confused again and now I remember why I found it so difficult to get started with this whole business.

mathause:
Yes FAIR (I think - at least one of the SCMs) provides smooth time series, so Lea smoothed it out and added artificial noise.

veni-vidi-vici-dormivi:
> provides smooth time series, so Lea smoothed it out

Why does she need to smooth it when it is already smooth?

As far as I can tell MAGICC provides smooth timeseries but with volcanic forcing already imposed.

mathause:
Ah sorry she smooths the GCM global mean time series, not the forcing one from the FAIR/ MAGICC.

PR #548, opened 2024-10-31
CI: test upstream dev

mathause:
- [ ] Closes #xxx
 - [ ] Tests added

This PR tests the development versions of our dependencies. This gives us more time if they introduce a breaking change and also helps to catch bugs in upstream dependencies. We don't update all the (indirect) dependencies (e.g. regionmask would depend on geopandas). The approach here is to install the current version, then remove it and install it with pip. The first step is done so we have all dependencies present.

That usually goes wrong the first time...

### DIFF
--- a/.github/workflows/upstream-dev-ci.yaml
+++ b/.github/workflows/upstream-dev-ci.yaml
@@ -1,0 +1,73 @@
+name: CI Upstream
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+    types: [opened, reopened, synchronize, labeled]
+  schedule:
+    - cron: "0 0 * * *" # Daily “At 00:00” UTC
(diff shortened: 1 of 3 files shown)


mathause:
Ok it finally runs. The failure is pandas 3 making indices read-only. Not sure how happy I am about this... (it's the same as mathause/mplotutils#133)

veni-vidi-vici-dormivi (approved):
Cool thanks!

mathause:
Thanks. So this runs daily with a cron job (we could run it less often...) and you can manually enable it for a PR by adding the `run-upstream` label.

veni-vidi-vici-dormivi:
> Thanks. So this runs daily with a cron job (we could run it less often...) and you can manually enable it for a PR by adding the `run-upstream` label.

Thanks!

PR #556, opened 2024-11-08
add Datatree utility functions

veni-vidi-vici-dormivi:
I add there utility functions to handle `DataTree`s for mesmer:
1. `_extract_single_dataarray_from_dt`: A lot of mesmer functions work on / return `xr.DataArray`s. This is suboptimal for the use of `DataTree` as they can only hold `xr.Dataset`s. So here is a function that extracts from a DataTree that holds only one Dataset with one DataArray this single DataArray.
2. `collapse_datatree_into_dataset` for the linear regression we want to stack all the scenarios into one DataArray. When we use a DataTree the different scenarios will be held in leaves/subtrees. This function collapses all subtrees of a DataTree into a Dataset along a given dimension (like 'scenario'.
3. For the linear regression using multiple members and scenarios we need to do some broadcasting, aligning and stacking. This is what `stack_linear_regression_datatrees` is for. Could also move this part to the PR dealing with the linear regression though...

Towards #106 
 - [x] Tests added


veni-vidi-vici-dormivi:
Okay here is the problem: if we want to use `datatree` (standalone) we can only use `v0.0.13` and need to keep the `xarray` version below `2024.10`. If we want to use `xr.DataTree` we need to push our minimum xarray version to `2024.10` or likely higher. I started working with `2024.10` a bit and it still has a few kinks that I am not happy with, mostly [#9693](https://github.com/pydata/xarray/issues/9693) and some missing functionality like `DataTree.broadcast_like()`...

veni-vidi-vici-dormivi:
In favor of time constraints it might be worth constraining our xarray version to between our minimum version and v2024.10 until we are happy with the xarray internal DataTree...

mathause:
That's annoying.

- `broadcast_like` is probably `map_over_subtree(xr.Dataset.broadcast_like)`
- do we have nodes without datasets? Is that the parent node `"/"`? - agreed that should be fixed
- I quickly checked #537 - all usages of `map_over_subtree` call internal mesmer functions so we could "fix" those

---

I am a bit hesitant to invest time into a package that is deprecated but given it works with datatree let's pin xarray & I'll worry about compatibility later.

veni-vidi-vici-dormivi:
Exactly, the parent note is always empty so most of my uses of map_over_subtree are invalid, including my fix for broadcast_like.

> I'll worry about compatibility later.

It think it's the best way to go at the moment and I already saw some things that are new in the `xr.DataTree` and am trying to already implement some of it.

mathause:
Ah upstream-dev-ci will not work (we can decide if we don't want to update xarray or disable it). It's not the most important thing but still. Also I don't like upper pins. But yes let's, pin, get that through and then fix compatibility.

veni-vidi-vici-dormivi:
Yay, thanks @mathause

mesmer/core/datatree.py
```diff
@@ -0,0 +1,160 @@
...
+    if not all(set(ds.dims) == first_dims for ds in datasets):
+        raise ValueError("All datasets must have the same dimensions")
+
+    # Check that all dimensions have coordinates
+    for ds in datasets:
+        for ds_dim in ds.dims:
+            if ds[ds_dim].coords == {}:
+                raise ValueError(
+                    f"Dimension '{ds_dim}' must have a coordinate/coordinates."
+                )
```
mathause, line 62:
```suggestion
        missing_coords = set(ds.dims) - set(ds.coords)
        if missing_coords:
            missing_coords = "', '".join(sorted(missing_coords))
            raise ValueError(
                f"Dimension(s) '{missing_coords}' must have a coordinate/coordinates."
            )
```

mesmer/core/datatree.py
```diff
@@ -0,0 +1,160 @@
...
+
+    # Concatenate datasets along the specified dimension
+    ds = xr.concat(datasets, dim=dim)
+    ds = ds.assign_coords(
+        {dim: [subtree.name for subtree in dt.subtree if not subtree.is_empty]}
```
mathause, line 67: Move name generation out of function call?

mesmer/core/datatree.py
```diff
@@ -0,0 +1,160 @@
...
+        leaf, holding a ``xr.Dataset`` representing a scenario. The subtrees of
+        different predictors must be isomorphic (i.e. have the save scenarios). The ``xr.Dataset``
+        must at least contain `dim` and each ``xr.Dataset`` must only hold one data variable.
+    target : DataTree
+        A ``DataTree``holding the targets. Must be isomorphic to the predictor subtrees, i.e.
+        have the same scenarios. Each leaf must hold a ``xr.Dataset`` which must be at least 2D
+        and contain `dim`, but may also contain a dimension for ensemble members.
+    weights : DataTree, default: None.
+        Individual weights for each sample, must be isomorphic to target. Must at least contain
+        `dim`, and must have the ensemble member dimesnion if target has it.
```
mathause, line 102:
```suggestion
        `dim`, and must have the ensemble member dimension if target has it.
```

mesmer/core/datatree.py
```diff
@@ -0,0 +1,160 @@
...
+    1. Broadcasts predictors to target
+    2. Collapses DataTrees into DataSets
+    3. Stacks the DataSets along the stacking dimension(s)
+
+    Parameters
+    ----------
+    predictors : DataTree
+        A ``DataTree`` of ``xr.Dataset`` objects used as predictors. The ``DataTree``
+        must have subtrees for each predictor each of which has to have at least one
+        leaf, holding a ``xr.Dataset`` representing a scenario. The subtrees of
```
mathause, line 93:
```suggestion
        must have subtrees for each predictor each of which has to have at least one
        non-empty leaf, representing a scenario. The subtrees of
```

maybe, not sure about this

mesmer/core/datatree.py
```diff
@@ -0,0 +1,160 @@
...
+    -----
+    Dimensions which exist along the target but are not in the stacking_dims will be excluded from the
+    broadcasting of the predictors.
+    """
+
+    stacking_dims_all = stacking_dims + [collapse_dim]
+    stack_dim = {stacked_dim: stacking_dims_all}
+
+    # exclude target dimensions from broadcasting which are not in the stacking_dims
+    exclude_dim = set(target.leaves[0].ds.dims) - set(stacking_dims)
```
mathause, line 128: Is this required to be non-empty? (But probably ok to defer the check to the linear regression function).

mesmer/core/datatree.py
```diff
@@ -0,0 +1,160 @@
...
+
+    # predictors need to be
```
mathause, line 130: Finish comment?

mesmer/core/datatree.py
```diff
@@ -0,0 +1,160 @@
...
+) -> tuple[DataTree, xr.Dataset, xr.Dataset | None]:
+    """
+    prepares data for Linear Regression:
+    1. Broadcasts predictors to target
+    2. Collapses DataTrees into DataSets
+    3. Stacks the DataSets along the stacking dimension(s)
+
+    Parameters
+    ----------
+    predictors : DataTree
```
mathause, line 90: Would it make sense to add some ASCII art to the notes? To illustrate how the predictors should look like? 

```
├─ tas
│  ├─ hist
│  └─ ...
├─ hfds
│  ├─ hist
│  └─ ...
└─ ...
```

mesmer/core/datatree.py
mathause: oh now I get it - maybe `# prepare predictors`? but also ok to keep

tests/unit/test_datatree.py
```diff
@@ -0,0 +1,263 @@
...
+
+def test_collapse_datatree_into_dataset():
+    n_ts = 30
+    ds1 = xr.Dataset({"tas": trend_data_1D(n_timesteps=n_ts)})
+    ds2 = ds1 * 2
+    ds3 = ds1 * 3
+
+    leaf1 = xr.concat([ds1, ds2, ds3], dim="member").assign_coords(
+        {"member": np.arange(3)}
+    )
```
mathause, line 19:
```suggestion
    dim = xr.Variable("member", np.arange(3))
    leaf1 = xr.concat([ds1, ds2, ds3], dim=dim)
```

mathause (commented):
I have not yet looked at all the tests but looks good overall!

mesmer/core/datatree.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+def collapse_datatree_into_dataset(
+    dt: DataTree, dim: str, **concat_kwargs
+) -> xr.Dataset:
+    """
+    Take a ``DataTree`` and collapse **all subtrees** in it into a single ``xr.Dataset`` along dim.
+    All subtrees are converted to ``xr.Dataset`` objects and concatenated along the
+    specified dimension. The dimension along which the datasets are concatenated will be added as a
+    coordinate to the resulting dataset and name of each subtree will be used as the coordinate
```
mathause, line 28:
```suggestion
    coordinate to the resulting dataset and the name of each subtree will be used as the coordinate
```

mesmer/core/datatree.py
```diff
@@ -0,0 +1,156 @@
...
+
+    da = ds[name]
+    return da
+
+
+def collapse_datatree_into_dataset(
+    dt: DataTree, dim: str, **concat_kwargs
+) -> xr.Dataset:
+    """
+    Take a ``DataTree`` and collapse **all subtrees** in it into a single ``xr.Dataset`` along dim.
```
mathause, line 25:
```suggestion
    Take a ``DataTree`` and collapse **all its subtrees** into a single ``xr.Dataset`` along dim.
```

mesmer/core/datatree.py
```diff
@@ -0,0 +1,156 @@
...
+
+    # Concatenate datasets along the specified dimension
+    ds = xr.concat(datasets, dim=dim, **concat_kwargs)
+    dims = [subtree.name for subtree in dt.subtree if not subtree.is_empty]
+    ds = ds.assign_coords({dim: dims})
+
+    return ds
+
+
+def stack_linear_regression_datatrees(
```
mathause, line 57:
```suggestion
def stack_datatrees_linear_regression(
```

?

or `stack_datatrees_for_linear_regression`

tests/unit/test_datatree.py
```diff
@@ -0,0 +1,272 @@
...
+    assert collapse_dim in res.dims
+    assert (res[collapse_dim] == ["scen1", "scen2"]).all()
+    assert len(res.dims) == 3
+    assert np.isnan(res.sel(scenario="scen2", member=2)).all()
+
+    # error if data set has no coords along dim (bc then it is not concatenable if lengths differ)
+    leaf_missing_coords = leaf1.drop_vars("member")
+    dt = DataTree.from_dict({"scen1": leaf_missing_coords, "scen2": leaf2})
+    with pytest.raises(
+        ValueError, match=("cannot reindex or align along dimension 'member'")
```
mathause, line 37:
```suggestion
        ValueError, match="cannot reindex or align along dimension 'member'"
```

tests/unit/test_datatree.py
```diff
@@ -0,0 +1,272 @@
...
+    ):
+        res = mesmer.datatree.collapse_datatree_into_dataset(dt, dim=collapse_dim)
+
+    # Dimension along which to concatenate already exists
```
mathause, line 41: nice test

tests/unit/test_datatree.py
```diff
@@ -0,0 +1,272 @@
...
+        shape=(n_lat * n_lon, n_samples),
+    )
+    _check_dataarray_form(
+        weights_stacked.weights, ndim=1, required_dims={"sample"}, shape=(n_samples,)
+    )
+
+    # check if datasets align
+    assert xr.align(
+        target_stacked, predictors_stacked["pred1"].to_dataset(), join="exact"
+    ) == (target_stacked, predictors_stacked["pred1"].to_dataset())
```
mathause, line 237: Do this in two steps? What does the `==` do for datasets? Ist that the same as `assert_equal`?

mathause (approved):
Thanks!

mathause:
We should maybe re-visit #348 to streamline where the data processing functions are exposed.

mesmer/core/datatree.py
veni-vidi-vici-dormivi: Ah okay thanks.

PR #562, opened 2024-11-12
tests: run in parallel

mathause:
- [ ] Closes #xxx
 - [ ] Tests added

Runs the tests in parallel. Can be called as

```bash
pytest -n auto ... # auto select number of cores
pytest -n 4 ... # set number of cores

```

While in theory very nice it does actually not speed up the tests locally. This surprises me but is maybe because the individual tests take a bit longer due to multi-threading issues. Let's see what happens on GHA. For some reason pytest-xdist is already in setup.cfg.

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -71,6 +71,7 @@ jobs:
 
     - name: Run tests
       run: python -u -m pytest
+       -n 4
        -r a -v
        --all
        --cov=mesmer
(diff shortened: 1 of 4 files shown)


mathause:
**single core**

image

**n=4**

image

veni-vidi-vici-dormivi (approved):
Sure, why not

mathause:
Thanks, yes I am a bit undecided but it's easy to revert so in it goes.

PR #566, opened 2024-11-21
Add data tree and dataset formats to linear regression

### LINKED ISSUE #355: stack predictors and targets for xarray objects
In #143 I restructured how predictors and targets are stacked. We need to do the same for `DataArray` objects.

- this will depend on how we structure our data
- I have a prototype of this based on `xr.broadcast`, `ds.stack`, and `xr.concat`

veni-vidi-vici-dormivi:
I implement handling of `DataTree`s and `xr.Dataset`s as predictor formats.
 - [x] Closes #355 
 - [x] Tests added


tests/unit/test_linear_regression.py
```diff
@@ -79,24 +93,74 @@ def test_lr_params():
 
 
 @pytest.mark.parametrize("as_2D", [True, False])
-def test_lr_predict(as_2D):
+@pytest.mark.parametrize("data_structure", [to_dict, to_datatree, to_xr_dataset])
```
mathause, line 96: Good approach!

tests/unit/test_linear_regression.py
```diff
@@ -79,24 +93,74 @@ def test_lr_params():
...
+        data_vars={
+            "intercept": ("x", [5]),
+            "fit_intercept": True,
+            "tas": ("x", [3]),
+            "tas2": ("x", [1]),
+        }
+    )
+    lr.params = params if as_2D else params.squeeze()
+
+    tas = xr.DataArray([0, 1, 2], dims="time").rename("tas")
```
mathause, line 129:
```suggestion
    tas = xr.DataArray([0, 1, 2], dims="time", name="tas")
```

tests/unit/test_linear_regression.py
mathause: Actually it's less clear in the code with the `data_structure`. Maybe rename to `to_data_type`?. Or maybe `data_type` is too close to `dtype` - could use `data_cls`?

<details>

Alternatively you could write

```python
def convert_to(dct, data_type):
    if data_type == "dict":
        return dct
    ...

# and 
@pytest.mark.parametrize("data_type", ["dict", "datatree", "dataset"])
def test_(...):
    ...

    pred = convert_to({"tas": tas}, data_type)

```

</details>

mathause (commented):
Looks good - some suggestions.

mesmer/stats/_linear_regression.py
```diff
@@ -88,30 +98,47 @@ def predict(
         if available_predictors - required_predictors:
             superfluous = sorted(available_predictors - required_predictors)
             superfluous = "', '".join(superfluous)
-            raise ValueError(f"Superfluous predictors: '{superfluous}'")
+            raise ValueError(
+                f"Superfluous predictors: '{superfluous}', either params",
+                "for this predictor are missing or you forgot to add it to 'exclude'.",
```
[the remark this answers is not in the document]
veni-vidi-vici-dormivi: 🥰

mesmer/stats/_linear_regression.py
```diff
@@ -88,30 +98,47 @@ def predict(
...
+            )
 
         if "intercept" in exclude:
             prediction = xr.zeros_like(params.intercept)
         else:
             prediction = params.intercept
 
+        # if predictors is a DataTree, rename all data variables to "pred" to avoid conflicts
+        if isinstance(predictors, DataTree) and not predictors.equals(DataTree()):
```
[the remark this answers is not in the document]
veni-vidi-vici-dormivi: No because is_empty only checks the node, so root, which can be empty while there are other datasets in the datatree. And we still need to check if it is all empty for the test without any predictors.

mesmer/stats/_linear_regression.py
```diff
@@ -182,12 +209,12 @@ def to_netcdf(self, filename, **kwargs):
             Additional keyword arguments passed to ``xr.Dataset.to_netcf``
         """
 
-        params = self.params()
+        params = self.params
```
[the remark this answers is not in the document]
veni-vidi-vici-dormivi: 🤗🤗

mesmer/stats/_linear_regression.py
```diff
@@ -88,30 +98,47 @@ def predict(
...
 
         if "intercept" in exclude:
             prediction = xr.zeros_like(params.intercept)
         else:
             prediction = params.intercept
 
+        # if predictors is a DataTree, rename all data variables to "pred" to avoid conflicts
+        if isinstance(predictors, DataTree) and not predictors.equals(DataTree()):
+            predictors = map_over_subtree(
+                lambda ds: ds.rename({var: "pred" for var in ds.data_vars})
```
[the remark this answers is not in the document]
veni-vidi-vici-dormivi: 🙃

tests/unit/test_linear_regression.py
veni-vidi-vici-dormivi: I like the one function with the string

mesmer/stats/_linear_regression.py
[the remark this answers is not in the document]
veni-vidi-vici-dormivi: We knew, from the `check_data_array` but I did not think of the case where dt only has data in root, which might be a use case if one has only one predictor. In this case, we now extract the dataset and treat it as such. For predict this should not be a problem.

veni-vidi-vici-dormivi:
Okay it is good to go from my side. The only thing that is a little sad is that one cannot use a `DataTree` or `Dataset` to make predictions with a coordinate like `time` but in this case one just has to pass it as a `dict`. 🤷‍♀️

mesmer/stats/_linear_regression.py
```diff
@@ -88,30 +98,45 @@ def predict(
...
             prediction = xr.zeros_like(params.intercept)
         else:
             prediction = params.intercept
 
+        # if predictors is a DataTree, rename all data variables to "pred" to avoid conflicts
+        # not necessaey if predictors is empty DataTree or only data is in root, i.e. depth == 0
+        if isinstance(predictors, DataTree) and not predictors.depth == 0:
+            predictors = map_over_subtree(
+                lambda ds: ds.rename({var: "pred" for var in ds.data_vars})
+            )(predictors)
```
mathause, line 116:
```suggestion
            (var,) = ds.data_vars
            predictors = map_over_subtree(
                lambda ds: ds.rename({var: "pred"})
            )(predictors)
```
to make it more obvious that it's only one variable?

mesmer/stats/_linear_regression.py
```diff
@@ -229,14 +261,32 @@ def _fit_linear_regression_xr(
...
+        # rename all data variables to "pred" to avoid conflicts when concatenating
+        predictors = map_over_subtree(
+            lambda ds: ds.rename({var: "pred" for var in ds.data_vars})
+        )(predictors)
+        predictors_concat = collapse_datatree_into_dataset(
+            predictors, dim="predictor", join="exact", coords="minimal"
+        )
+        predictors_concat = (
+            predictors_concat.to_array().isel(variable=0).drop_vars("variable")
+        )
```
mathause, line 289: Can you explain me why you have to go via `to_array`? Also please use `to_dataarray` if it cannot be avoided (which is the same as `to_array` but with a better name).

tests/unit/test_datatree.py
```diff
@@ -60,6 +60,17 @@ def test_collapse_datatree_into_dataset():
 
     xr.testing.assert_equal(scen1.drop_vars(collapse_dim), leaf1)
 
+    # data in root works
```
mathause, line 63:
```suggestion
    # test data in root works
```

tests/unit/test_linear_regression.py
```diff
@@ -4,11 +4,24 @@
...
 import pytest
 import xarray as xr
+from datatree import DataTree
 
 import mesmer
+import mesmer.stats._linear_regression
 from mesmer.testing import trend_data_1D, trend_data_2D
 
 
+def convert_to(dct: dict, data_type: str) -> dict | DataTree | xr.Dataset:
```
mathause, line 14: :+1:

mathause (approved):
Two more questions but good to merge. Thanks!

mesmer/create_emulations/utils.py
```diff
@@ -114,7 +114,7 @@ def _gather_lr_params(params_dict, targ, dims):
         intercept = xr.DataArray(params_dict["intercept"][targ], dims=dims)
         fit_intercept = True
     else:
-        intercept = 0
+        intercept = xr.zeros_like(params[pred])
```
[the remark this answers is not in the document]
veni-vidi-vici-dormivi: Hm apparently it is not... Maybe it never was. I had a problem I think because the prediction will have the shape of the intercept which might have lead to problems in `(predictors[key] * params[key]).transpose() + prediction`, but maybe I used a different order of operations then. Now it seems to work. Thanks!

mesmer/stats/_linear_regression.py
veni-vidi-vici-dormivi: ds is not defined outside of map_over_subtree sadly, one could do `ds.rename({list(ds.data_vars)[0]: "pred"})`

mesmer/stats/_linear_regression.py
veni-vidi-vici-dormivi: Ah we can also do `predictors_concat['pred']`

mesmer/stats/_linear_regression.py
mathause: Ah yes, or go with a function instead of a lambda. I'll leave it up to you

PR #568, opened 2024-11-27
show timing of slowest tests

mathause:
- [ ] Closes #xxx
 - [ ] Tests added

Shows the timing of the 10 slowest tests each time the tests are run. I wasn't sure if I should only add this for the CI :man_shrugging: (i.e. add to the pytest call in ci-workflow.yml)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -101,7 +101,7 @@ omit =
     */tests/*, */scripts/*, */examples/*, */configs/*, */mesmer/_version.py
 
 [tool:pytest]
-addopts = --strict-markers
+addopts = --strict-markers --durations=10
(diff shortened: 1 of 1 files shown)


veni-vidi-vici-dormivi:
Wdym "only"? Where else would you add it? I think it's cool!

mathause:
Now it also shows when you run the tests locally which takes up some lines on your terminal but yes I think that is fine. The alternative is to add `--durations=10` in

https://github.com/MESMER-group/mesmer/blob/9ebcd27813ece4eae03aedd220820150b3a3ce94/.github/workflows/ci-workflow.yml#L73

PR #574, opened 2024-12-04
Let fit_harmonic_model return residuals instead of predictions

veni-vidi-vici-dormivi:
This way we can avoid that the user has to compute the residuals manually. We never need the predictions, only the residuals, so this makes sense in any case but it is also nice because it a) prevents users (ehem *me*) from mixing up the order (`resids = predictions - target` or `target - predictions`, which leads to different `lambdas`) and b) is practical for the use of `DataTree`. 

I also drop the residuals from the fitted harmonic model and AR parameter sets to save some space as we do not need them for emulation. 

 - [x] Tests adjusted


CHANGELOG.rst
```diff
@@ -177,8 +177,9 @@ Harmonic model
    (`#415 <https://github.com/MESMER-group/mesmer/pull/415>`_,
    `#424 <https://github.com/MESMER-group/mesmer/pull/424>`_,
    `#433 <https://github.com/MESMER-group/mesmer/pull/433>`_,
+   `#512 <https://github.com/MESMER-group/mesmer/pull/512>`_,
```
mathause, line 180:
```suggestion
```

That seems to be listed twice?

mesmer/stats/_harmonic_model.py
```diff
@@ -267,7 +267,7 @@ def fit_harmonic_model(yearly_predictor, monthly_target, max_order=6, time_dim="
     data_vars : `xr.Dataset`
         Dataset containing the selected order of Fourier Series (selected_order),
         the estimated coefficients of the Fourier Series (coeffs) and the resulting
-        predictions for monthly values (predictions).
+        residuals of th model (residuals).
```
mathause, line 270:
```suggestion
        residuals of the model (`residuals`).
```

Enclose all variable names in backticks?

mathause (approved):
Yes fair enough. Does the notebook need an update?

veni-vidi-vici-dormivi:
> Yes fair enough. Does the notebook need an update?

Oh yes, thank you!

PR #578, opened 2024-12-09
harmonic model: more performance optimization

mathause:
- [ ] Closes #xxx
 - [ ] Tests added

Found another couple of performance optimizations. The most important is caching `cos_sin`. I knew that last time but did not think of using an `lru_cache` which makes the second call almost free (`50 ns` vs `10 μs` for `order=6`; which equals about 1/3 of the total time of `_generate_fourier_series_np`). Also not truncating coeffs (when known not to contain `nan`) and using `ravel` brings another ~5% each. However, it's not super visible in the timing of the tests...


veni-vidi-vici-dormivi (approved):
Wow very cool, thanks!

Issue #525, opened 2024-09-18
`Expression` fails with extra commas


mathause:
I am writing unit tests for `Expression` and it fails if there are more commas than parameters, as it splits on them.

```python
from mesmer.mesmer_x import Expression

mesmer.mesmer_x.Expression("norm(loc=c1, scale=np.mean([c2, c3])", "name")
```

```python-traceback
ValueError                                Traceback (most recent call last)
Cell In[38], line 1
----> 1 mesmer.mesmer_x.Expression("norm(loc=c1, scale=np.mean([c2, c3])", "name")

File ~/code/mesmer/mesmer/mesmer_x/train_utils_mesmerx.py:94, in Expression.__init__(self, expr, expr_name)
     92 # identify parameters
     93 self.find_parameters_list()
---> 94 self.find_expr_parameters()
     96 # identify coefficients
     97 self.find_coefficients()

File ~/code/mesmer/mesmer/mesmer_x/train_utils_mesmerx.py:134, in Expression.find_expr_parameters(self)
    132 self.parameters_expressions = {}
    133 for sub_exp in sub_expressions:
--> 134     param, sub = str.split(sub_exp, "=")
    135     if param in self.parameters_list:
    136         self.parameters_expressions[param] = sub

ValueError: not enough values to unpack (expected 2, got 1)
```

mathause:
I hit comment to early. cc @yquilcaille - I am not sure we should try to fix this but think about if there is a different approach we can take. Such a parsing is inherently difficult...

yquilcaille:
My naive opinion would be simply to change the character for the parsing, and use something that would not be used normally in python. Maybe $? But i see the issue [(#527)](https://github.com/MESMER-group/mesmer/issues/527), see whether it is a better fix.

mathause:
I would like to avoid inventing our own language :wink: Instead we could

1. split on `=` (although that may fail on functions which require params, e.g. `axis=0`)
1. use regex but that would be fiendishly difficult (so :-1: from my side)
1. require each param to have it's own keyword argument, e.g.:
   `Expression("norm", loc="c1", scale="np.mean([c2, c3])", name="name")`
   - but would have to use `**kwargs` if we keep the generic `Expression` (but again we are in the same situation today: the user needs to know the required params for the distribution)

mathause:
Two more issues:

1. coefficients can only have one digit:
   ```python
[...]

PR #597, opened 2025-01-23
Enable that location and scale parameters might be set or not used.

veni-vidi-vici-dormivi:
Some distributions do not use `loc` and `scale`. While all distributions have `loc` as parameter, for some it might make sense to set it to 0 like the beta distribution or even just if one has calculated residuals for their data and knows the loc is 0. Thus, I implemented an option to set the location in the expression to a certain value (note because of #525 it can only be ints atm). Moreover, the discrete distributions tend to not have a scale parameter so I made it optional.

 - [x] Tests added

### DIFF
--- a/mesmer/mesmer_x/train_l_distrib_mesmerx.py
+++ b/mesmer/mesmer_x/train_l_distrib_mesmerx.py
@@ -1042,14 +1042,17 @@ class distrib_cov:
             [self.expr_fit.coefficients_list.index(c) for c in loc_coeffs]
         )
 
-        localfit_loc = self._minimize(
-            func=self._fg_fun_loc,
-            x0=self.fg_coeffs[self.fg_ind_loc],
-            args=(smooth_targ,),
-            fact_maxfev_iter=len(self.fg_ind_loc) / self.n_coeffs,
-            option_NelderMead="best_run",
-        )
-        self.fg_coeffs[self.fg_ind_loc] = localfit_loc.x
+        # location might not be used (beta distribution) or set in the expression
+        if len(self.fg_ind_loc) > 0:
+
+            localfit_loc = self._minimize(
+                func=self._fg_fun_loc,
+                x0=self.fg_coeffs[self.fg_ind_loc],
+                args=(smooth_targ,),
+                fact_maxfev_iter=len(self.fg_ind_loc) / self.n_coeffs,
+                option_NelderMead="best_run",
+            )
+            self.fg_coeffs[self.fg_ind_loc] = localfit_loc.x
 
         # Step 3: fit coefficients of scale (objective: improving the subset of
         # scale coefficients)
@@ -1058,7 +1061,7 @@ class distrib_cov:
             scale_coeffs = self.expr_fit.coefficients_dict["scale"]
         except KeyError: 
             scale_coeffs = []
-        
+        # scale might not be used or set in the expression
         if len(scale_coeffs) > 0:
             self.fg_ind_sca = np.array(
                 [self.expr_fit.coefficients_list.index(c) for c in scale_coeffs]
--- a/mesmer/mesmer_x/train_l_distrib_mesmerx.py
+++ b/mesmer/mesmer_x/train_l_distrib_mesmerx.py
@@ -1061,9 +1061,10 @@ class distrib_cov:
             scale_coeffs = self.expr_fit.coefficients_dict["scale"]
         except KeyError:
             scale_coeffs = []
-            self.fg_ind_sca = np.array(
-                [self.expr_fit.coefficients_list.index(c) for c in scale_coeffs]
-            )
+        
+        self.fg_ind_sca = np.array(
+            [self.expr_fit.coefficients_list.index(c) for c in scale_coeffs]
+        )
         # scale might not be used or set in the expression
         if len(self.fg_ind_sca) > 0:
             if self.first_guess is None:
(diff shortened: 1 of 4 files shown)


veni-vidi-vici-dormivi:
Test for discrete distributions that actually omit the scale parameter follow in another PR as they also need some other changes.

mesmer/mesmer_x/train_l_distrib_mesmerx.py
```diff
@@ -1037,44 +1037,54 @@ def find_fg(self):
...
+                fact_maxfev_iter=len(self.fg_ind_loc) / self.n_coeffs,
+                option_NelderMead="best_run",
+            )
+            self.fg_coeffs[self.fg_ind_loc] = localfit_loc.x
 
         # Step 3: fit coefficients of scale (objective: improving the subset of
         # scale coefficients)
-        scale_coeffs = self.expr_fit.coefficients_dict["scale"]
+        # TODO: move to `Expression`
+        try:
```
yquilcaille, line 1060: I think that all scipy stats have the capacity to have a loc & scale. But it is true that the user may decide not to provide it. In this case, the default of scipy.stats will automatically be used, thus a scale of 1. And if the location is not provided, a default of 0.
Should we also add a "try / except KeyError" if the user decides not to provide a location? e.g. a beta with the parameters a and b varying, but nothing said about loc and scale?

mesmer/mesmer_x/train_l_distrib_mesmerx.py
```diff
@@ -1037,44 +1037,54 @@ def find_fg(self):
...
+            scale_coeffs = self.expr_fit.coefficients_dict["scale"]
+        except KeyError:
+            scale_coeffs = []
```
mathause, line 1063:
```suggestion
        scale_coeffs = self.expr_fit.coefficients_dict.get("scale", [])
```

tests/unit/test_mesmer_x_first_guess.py
```diff
@@ -144,6 +144,70 @@ def test_first_guess_truncnorm():
...
+        targ,
+        {"tas": pred},
+        expression,
+        first_guess=first_guess,
+        options_solver=options_solver,
+    )
+    dist.find_fg()
+
+    # NOTE: for the beta distribution the support does not change for loc = 0 and scale = 1
+    # it is always (0,1), thus the optimization with _fg_fun_others does not do anything
```
mathause, line 174:
```suggestion
    # it is always (0, 1), thus the optimization with _fg_fun_others does not do anything
```

mathause (commented):
Looks good.

mesmer/mesmer_x/train_l_distrib_mesmerx.py
veni-vidi-vici-dormivi: So all continuous distributions have loc and scale but the discrete ones have only loc. We check that the user provides information for these two in the expression:

https://github.com/MESMER-group/mesmer/blob/2208ee662e9c56fc5492cf0f692f883e13bc360f/mesmer/mesmer_x/train_utils_mesmerx.py#L168-L171

So from the expression it is not possible to omit loc or scale for a continuous distribution and not possible to omit loc for the discrete ones. That is why I made scale optional but not loc. Yes the user might sometimes need to provide a loc where it is not necessary but I think that is not so bad. I would leave it as is and maybe think about changing the expression later. What do you think?

mesmer/mesmer_x/train_l_distrib_mesmerx.py
veni-vidi-vici-dormivi: fancy, I like it

mesmer/mesmer_x/train_l_distrib_mesmerx.py
```diff
@@ -564,8 +564,8 @@ def __init__(
             "method_fit": "Powell",
             "xtol_req": 1e-6,
             "ftol_req": 1.0e-6,
-            "maxiter": 1000 * self.n_coeffs * np.log(self.n_coeffs),
-            "maxfev": 1000 * self.n_coeffs * np.log(self.n_coeffs),
+            "maxiter": 1000 * self.n_coeffs * (np.log(self.n_coeffs) + 1),
+            "maxfev": 1000 * self.n_coeffs * (np.log(self.n_coeffs) + 1),
```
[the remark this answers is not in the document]
veni-vidi-vici-dormivi: thanks :)

mesmer/mesmer_x/train_l_distrib_mesmerx.py
yquilcaille: True for the discrete distributions. However, the location can still be omitted. For instance, we can write `scipy.stats.beta(a=1, b=2).rvs()`, so an user could proposer an expression without location. So an user could propose the expression to train `scipy.stats.beta(a=c1, b=c2 + c3 * __GMT__)`. Do we authorize that or would this possibility not make sense?

veni-vidi-vici-dormivi: We don't allow it (yet)

https://github.com/MESMER-group/mesmer/blob/4ef7a1f25d11d60ea5c84e6dde7eb472c3b2756e/mesmer/mesmer_x/train_utils_mesmerx.py#L144-L148

Since we always add loc to the parameters list it will throw and error here in `Expression._find_expr_parameters()`

yquilcaille: True. Then good to go!

veni-vidi-vici-dormivi: Thanks! :)

mesmer/mesmer_x/train_l_distrib_mesmerx.py
veni-vidi-vici-dormivi: I know you showed me this before, but it needs some more time until it is moved to my active knowledge base 😅

Issue #276, opened 2023-09-04
deprecate old-style functions


mathause:
We should raise deprecations warnings (well `FutureWarning`...)  for the legacy functions. I am not 100% sure if already for 0.9 or only in 1.0. Also ensure the warnings are not raised 100 times...

mathause:
Moving this to the v0.10.0 milestone

mathause:
copied from https://github.com/MESMER-group/mesmer/pull/599#issuecomment-2615705792 by @veni-vidi-vici-dormivi 

> > Yes agree to deprecate the rest - I am not sure what the best approach is, though - maybe a decorator?
> 
> Hm as far as is see the decorator needs a new dependency?

mathause:
I don't think so - or what did you have in mind? I thought along the lines of

```python

from functools import wraps
import warnings

def deprecate(func):

    @wraps(func)
    def wrapped(*args, **kwargs):
        warnings.warn(
            "This mesmer code path is deprecated. Please use the new api. See example"
            " at XXX"
            FutureWarning
        )

        func(*args, **kwargs)

    return wrapped
```

We could also take a look at https://github.com/matplotlib/matplotlib/blob/main/lib/matplotlib/_api/deprecation.py but that is probably overkill

veni-vidi-vici-dormivi:
Ah I thought you meant https://deprecated.readthedocs.io/en/latest/introduction.html

But your solution seems to do the same thing.
But, this would mean all the functions are still there right? So also the tests have to stay. (I was hoping we could get rid of some of the old test data, but its also fine if not)

mathause:
Ah didn't know about this package. It seems to be a similar approach than what I suggested (a bit more fancy and also for classes and awaitable functions etc.). We can also just remove the old code path. Or remove the tests and exclude them from the coverage (e.g. in combination with #262). We could also just raise a `DeprecationError` if one of the old style functions are imported or used. One thing I want to check before we do: did we do any changes to the old code path since the last release. Or yet another solution: make a release, remove the old style functions and do another release quickly after that (only question: which one is v1.0?)

Matlab also has a bunch of helper classes https://github.com/matplotlib/matplotlib/blob/main/lib/matplotlib/_api/deprecation.py where we could draw inspiration from (but probably also not necessary).

veni-vidi-vici-dormivi:
So the only thing we changes is the switch from averaging standard deviations to averaging variances. 

[...]

PR #599, opened 2025-01-23
Add new emulation integration tests

veni-vidi-vici-dormivi:
I added emulation tests for the new code path also using `DataTree`. I added also tests to draw realizations for several forcing scenarios at once, which is new compared to the old tests. What is also new is that I actually use scenario data as forcing data while the old tests used dummy data. There is no real advantage of that I think, I just liked the idea of the test being consistent with the verification process. Note that when I draw realizations for ssp585 and ssp126 I also use the parameter set that was calibrated on those two scenarios which is not really necessary. However, this implicitly tests that all parameter sets we created lead to stable emulations. 


veni-vidi-vici-dormivi:
Okay this would maybe be a good time for #262 respectively #276. If we deprecate the old functions, could we remove the tests without loosing coverage @mathause? It would be nice we could delete the bundle and the old emulation output files from the repo to save some space.

tests/integration/test_draw_realisations_mesmer_newcodepath.py
```diff
@@ -0,0 +1,372 @@
...
+import mesmer
+
+
+# comment
+def create_forcing_data(test_data_root_dir, scenarios, use_hfds, use_tas2):
+    # define config values
+    REFERENCE_PERIOD = slice("1850", "1900")
+
+    esm = "IPSL-CM6A-LR"
+    test_cmip_generation = 6
```
mathause, line 18:
```suggestion
    cmip_generation = 6
```
?

tests/integration/test_draw_realisations_mesmer_newcodepath.py
```diff
@@ -0,0 +1,372 @@
...
+
+    # define paths and load data
+    TEST_DATA_PATH = pathlib.Path(test_data_root_dir)
+
+    cmip_data_path = (
+        TEST_DATA_PATH / "calibrate-coarse-grid" / f"cmip{test_cmip_generation}-ng"
+    )
+
+    CMIP_FILEFINDER = FileFinder(
+        path_pattern=cmip_data_path / "{variable}/{time_res}/{resolution}",  # type: ignore
```
mathause, line 28: Didn't we find out that this doesn't work? But now it does :thinking: well ok then

tests/integration/test_draw_realisations_mesmer_newcodepath.py
```diff
@@ -0,0 +1,372 @@
...
+    def _get_hist(meta, fc_hist):
+
+        meta_hist = meta | {"scenario": "historical"}
+
+        fc = fc_hist.search(**meta_hist)
+
+        if len(fc) == 0:
+            raise FileNotFoundError("no hist file found")
+        if len(fc) != 1:
+            raise ValueError("more than one hist file found")
```
mathause, line 58: mpytools/filefisher#150 (no TODO)

tests/integration/test_draw_realisations_mesmer_newcodepath.py
```diff
@@ -0,0 +1,372 @@
...
+
+        fN, meta_hist = fc[0]
+
+        return fN, meta_hist
```
mathause, line 62:
```suggestion
        return fc.paths[0]
```

?

tests/integration/test_draw_realisations_mesmer_newcodepath.py
```diff
@@ -0,0 +1,372 @@
...
+        variable="tas",
+        scenario="historical",
+        model=esm,
+        resolution="g025",
+        time_res="ann",
+        member=unique_scen_members,
+    )
+
+    # load data for each scenario
+    def _get_hist(meta, fc_hist):
```
mathause, line 49:
```suggestion
    def _get_hist_path(meta, fc_hist):
```
?

tests/integration/test_draw_realisations_mesmer_newcodepath.py
```diff
@@ -0,0 +1,372 @@
...
+    tas_anoms = tas - ref
+    tas_globmean = map_over_subtree(mesmer.weighted.global_mean)(tas_anoms)
+
+    tas_globmean_ensmean = tas_globmean.mean(dim="member")
+    tas_globmean_forcing = map_over_subtree(mesmer.stats.lowess)(
+        tas_globmean_ensmean, dim="time", n_steps=30, use_coords=False
+    )
+
+    if use_hfds:
+        fc_hfds = CMIP_FILEFINDER.find_files(
```
mathause, line 120: move this block to a function

tests/integration/test_draw_realisations_mesmer_newcodepath.py
```diff
@@ -0,0 +1,372 @@
...
+        module="local-trends", esm=esm, scen=scen_str
+    ).paths.pop()
+    local_ar_file = PARAM_FILEFINDER.find_single_file(
+        module="local-variability", esm=esm, scen=scen_str
+    ).paths.pop()
+    localized_ecov_file = PARAM_FILEFINDER.find_single_file(
+        module="covariance", esm=esm, scen=scen_str
+    ).paths.pop()
+
+    expected_output_file = TEST_PATH / "test_make_realisations_expected_output.nc"
```
mathause, line 257: should we add `outpath` to the file name?

tests/integration/test_draw_realisations_mesmer_newcodepath.py
```diff
@@ -0,0 +1,372 @@
...
+
+    HIST_PERIOD = slice("1850", "2014")
+
+    # we need a maximum of 4 seeds if there are max 2 scenarios (1 for global and 1 for local)
+    seed_list = [981, 314, 272, 42]
```
mathause, line 262: nice seeds :wink:

tests/integration/test_draw_realisations_mesmer_newcodepath.py
```diff
@@ -0,0 +1,372 @@
...
+
+    buffer_global_variability = 50
+    buffer_local_variability = 20
+
+    volcanic_params = xr.open_dataset(volcanic_file)
+    tas_forcing = map_over_subtree(mesmer.volc.superimpose_volcanic_influence)(
+        tas_forcing, volcanic_params, hist_period=HIST_PERIOD
+    )
+
+    # 2.) compute the global variability
```
mathause, line 289: Is there a 1?

mathause (commented):
Super! Yes agree to deprecate the rest - I am not sure what the best approach is, though - maybe a decorator?

tests/integration/test_draw_realisations_mesmer_newcodepath.py
veni-vidi-vici-dormivi: the pathlib.path? No we found out it does work I think because the regular expression part can handle paths and transforms them into strings?

tests/integration/test_draw_realisations_mesmer_newcodepath.py
veni-vidi-vici-dormivi: True, that would be nice.

tests/integration/test_draw_realisations_mesmer_newcodepath.py
veni-vidi-vici-dormivi: hihihi I was hoping you would say that 😄❤️

tests/integration/test_draw_realisations_mesmer_newcodepath.py
veni-vidi-vici-dormivi: ah I forgot the comment. yes its the volcanic influence

tests/integration/test_draw_realisations_mesmer_newcodepath.py
veni-vidi-vici-dormivi: hm it is already in the folder that denotes the `multi_scen_multi_ens` part for example but we could add the scen string, for example `ssp126-ssp585`

tests/integration/test_draw_realisations_mesmer_newcodepath.py
mathause: I thought it does not work but obviously fine if it does

veni-vidi-vici-dormivi:
> Yes agree to deprecate the rest - I am not sure what the best approach is, though - maybe a decorator?

Hm as far as is see the decorator needs a new dependency?

mathause (approved):
Thanks

PR #602, opened 2025-01-24
distrib_cov._fg_fun_LL_n(): add logpmf for discrete distributions

veni-vidi-vici-dormivi:
Discrete distributions don't have logpdf but instead a logpmf. We already use the logpmf in `distrib_cov.loglike()` but  we do not reuse this function here because we need an exponent. Thus I added the logpmf for discrete distribution which enables us to also test some of these.

 - [x] Tests added

### DIFF
--- a/tests/unit/test_mesmer_x_first_guess.py
+++ b/tests/unit/test_mesmer_x_first_guess.py
@@ -1,6 +1,6 @@
 import numpy as np
 import pytest
-from scipy.stats import beta, binom, gamma, hypergeom, genextreme, laplace, truncnorm
+from scipy.stats import beta, binom, gamma, genextreme, hypergeom, laplace, truncnorm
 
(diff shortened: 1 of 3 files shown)


tests/unit/test_mesmer_x_first_guess.py
```diff
@@ -150,6 +150,68 @@ def test_first_guess_truncnorm():
...
+    result = dist.fg_coeffs
+    expected = [n_trials, p]
+
+    np.testing.assert_allclose(result, expected, rtol=0.2)
+
+
+def test_fg_hypergeom():
+    rng = np.random.default_rng(0)
+    n = 251
+    pred = np.ones(n, dtype=int) * 2
```
mathause, line 176:
```suggestion
    pred = np.full(n, fill_value=2, dtype=int)
```

mathause (approved):
Thanks - looks good.

PR #607, opened 2025-01-31
compatibility with xr.DataTree

mathause:
- [ ] Closes #xxx
 - [ ] Tests added

Adds compatibility with `xr.DataTree`. Works (at least locally) but is not finished and needs some ugly workarounds due to some design decisions in the new version... The new code path is only tested in upstream-dev ci.

I am not sure we should support both, datatree and xarray datatree, but this was actually the smaller part of the whole operation.


mathause:
The upstream packages seem flaky

veni-vidi-vici-dormivi (approved):
Wow thanks, that is a lot of work! I think it's okay not to support `xarray-datatree` at all then, since it is not recommended to use it anyway.

tests/integration/test_calibrate_mesmer_newcodepath.py
```diff
@@ -183,13 +184,17 @@ def test_calibrate_mesmer(
...
     # convert the 0..360 grid to a -180..180 grid to be consistent with legacy code
 
     # calculate anomalies w.r.t. the reference period
     ref = dt["historical"].sel(time=REFERENCE_PERIOD).mean("time")
-    tas_anoms = dt - ref.ds
-    tas_globmean = map_over_subtree(mesmer.weighted.global_mean)(tas_anoms)
+
+    # https://github.com/pydata/xarray/issues/10013
+    # tas_anoms = dt - ref.ds
+    tas_anoms = map_over_datasets(operator.sub, dt, ref.ds)
```
mathause, line 195: This takes all the fun out of using `xr.DataTree` and cannot be expected from a user. The other 'issues' (missing methods) will probably be fixed in time but this seems to be a design decision. :-1:

tests/integration/test_calibrate_mesmer_newcodepath.py
```diff
@@ -234,12 +239,23 @@ def mask_and_stack(ds, threshold_land):
...
-        hfds_globmean_smoothed = map_over_subtree(mesmer.stats.lowess)(
-            hfds_globmean_ensmean, "time", n_steps=50, use_coords=False
+        hfds_globmean_smoothed = map_over_datasets(
+            mesmer.stats.lowess,
+            hfds_globmean_ensmean,
+            "time",
+            n_steps=50,
+            use_coords=False,
         )
+        print(hfds_globmean_smoothed)
```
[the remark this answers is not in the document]
mathause:
```suggestion
```

Thanks missed that.

mathause (commented):
Thanks. It was mostly annoying because 98 % I did not understand what is happening :man_shrugging:

As long as we need to patch `xr.map_over_datasets` we can just as well support datatree (because we need to import from _datatreecompat anyway.

It's annoying that the new version has become so bothersome to work with :unamused:

tests/integration/test_calibrate_mesmer_newcodepath.py
veni-vidi-vici-dormivi: Yes I agree.

mathause:
I have not yet updated the example scripts and we need to fix the pin for xarray (so it's not only tested in upstream-dev ci).

tests/integration/test_calibrate_mesmer_newcodepath.py
mathause: I think what we have to do is write a helper function `calc_anomaly(dt, ref_period)` with this inside.

mathause:
I opened pydata/xarray#10042 - there seems to be some willingness to enable skipping empty nodes. With this change we could avoid the `skip_empty_nodes` helper function and I'd say that's the moment we can "fully" support `xr.DataTree`.

But not super convinced this PR gets a fast turnaround :unamused:

mathause:
Ah yes and after 507cd1c this definitely requires xarray main, i.e. we require

```
xarray >=2023.07, < 2024.10, > 2025.1
```

(thats probably not a valid specifier)

yquilcaille:
Should we merge this branch? I'm quite blocked now, because I'm trying to use the former datatree, but knowing that we will use this PR...

yquilcaille:
Actually, there are no imports in the __init__.py of mesmer.core. Should we add a commit doing the following?

`# MESMER, land-climate dynamics group, S.I. Seneviratne
# Copyright (c) 2021 ETH Zurich, MESMER contributors listed in AUTHORS.
# Licensed under the GNU General Public License v3.0 or later see LICENSE or
# https://www.gnu.org/licenses/
"""
Collection of functions to load in various datasets.
"""
# flake8: noqa

from ._data import *
from .datatree import *
from .geospatial import *
from .grid import *
from .mask import *
from .regionmaskcompat import *
from .utils import *
from .volc import *
from .weighted import *`

mathause:
> Actually, there are no imports in the **init**.py of mesmer.core. Should we add a commit doing the following?
> 
> ```
> # flake8: noqa
> 
> from ._data import *
> from .datatree import *
> from .geospatial import *
> from .grid import *
> from .mask import *
> from .regionmaskcompat import *
> from .utils import *
> from .volc import *
> from .weighted import *
> ```

Users don't really need to access functions in `core` so there is no need to add these. Except for the ones in datatreecompat. We should make them top-level functions (`mesmer.DataTree`). (However, I would very much prefer that we don't need to release this and am still hoping for https://github.com/pydata/xarray#10042.)

mathause:
This is not moving forward on xarrays site, so I am merging. I will merge +- as is. Open a PR for a `calc_anomaly` helper function, then drop support for the old datatree package (i.e. require xarray >= 2025.02).

mathause:
I also have to find out if `map_over_datasets` is only used internally. If not I'll make it a top-level function...

mathause:
And we can always remove the compatibility `map_over_datasets` if that get's resolved before we release.

PR #613, opened 2025-02-03
MESMER-X: keep scalars

mathause:
- [ ] Closes #xxx
 - [ ] Tests added

I am very confident that the scalars don't need to be broadcast, which will save some more seconds.


mathause:
The speedup is not obvious but I saw it locally when profiling.

PR #617, opened 2025-02-06
Raise error if first element returns `inf` in `_minimize_local_discrete`

### LINKED ISSUE #616: Bug in `_minimize_local_discrete`

veni-vidi-vici-dormivi:
This is to avoid if the first element returns `inf`, the last element is returned. 

Note that for the application in `find_localized_empirical_covariance()` the whole thing implicitly assumes that the localization radii are ordered and increasing. And that once a localization radius leads to a singular covariance matrix all higher ones do too (reasonably).

 - [ ] Closes #616
 - [ ] Tests added

### DIFF
--- a/mesmer/core/utils.py
+++ b/mesmer/core/utils.py
@@ -43,12 +43,13 @@ def _minimize_local_discrete(func: Callable, sequence: Iterable, **kwargs):
 
     Raises
     ------
-    ValueError : if `func` returns negative infinity for any input.
+    ValueError : if `func` returns negative infinity for any input or first element is `inf`.
 
     Notes
     -----
     - The function determines the local minimum, i.e., the loop is aborted if
       `func(sequence[i-1]) >= func(sequence[i])`.
+    - If func returns `inf` for any input, the function will return the previous element.
     """
 
     current_min = float("inf")
@@ -73,9 +74,14 @@ def _minimize_local_discrete(func: Callable, sequence: Iterable, **kwargs):
             sel = i - 1
             if sel == 0:
                 warnings.warn("First element is local minimum.", OptimizeWarning)
+            elif sel == -1:
+                raise ValueError("First element is `inf`, aborting.")
+
             return sequence[sel]
 
-    warnings.warn("No local minimum found, returning the last element", OptimizeWarning)
+        warnings.warn(
+            "No local minimum found, returning the last element", OptimizeWarning
+        )
 
     return element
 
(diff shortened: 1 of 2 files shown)


veni-vidi-vici-dormivi:
Note also that we do not use `minimize_local_discrete` anywhere else at the moment.

veni-vidi-vici-dormivi:
Closing again in favor of #493

veni-vidi-vici-dormivi:
Okay leaving this here in case we decide to not merge #493

mathause:
Ok I only now really got the issue (probably). What if we used `res <= current_min` instead of `res < current_min`? We can then check if the last element is `inf`.

I pushed 2 commits with this change.

PR #624, opened 2025-03-24
fit_harmonic_model: more validity checks of input

### LINKED ISSUE #589: MESMER-M minor issue with harmonic model
Function fit_harmonic_model does not throw an error when calling it with predictor and target data that does not have the same spatial structure (e.g. calling it with tas_globmean as predictor). I'd suggest to add 

`def fit_harmonic_model(..., gridcell_dim = 'gridcell')`

and

```python
if not len(monthly_target[gridcell_dim]) == len(yearly_predictor[gridcell_dim]):
    raise ValueError("Yearly predictor values and monthly target data must have the same number of gridcells")
```

And there is also v minor error in the function description, I think it should be:

```
monthly_target : xr.DataArray of shape (**n_months*n_years**, n_gridcells)
```

mathause:
- [x] Closes #589
 - [x] Tests added

@sarasita thanks for the issue. I added some more sanity checks for the input of `fit_harmonic_model`

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -208,8 +208,9 @@ Harmonic model
    (`#415 <https://github.com/MESMER-group/mesmer/pull/415>`_,
    `#424 <https://github.com/MESMER-group/mesmer/pull/424>`_,
    `#433 <https://github.com/MESMER-group/mesmer/pull/433>`_,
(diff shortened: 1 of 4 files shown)


mathause:
this will still issue a not-nice error message if the time coordinates are not datetime like, but I did not find a simple way to check for this (pydata/xarray#10193)

PR #625, opened 2025-03-31
add anomaly function

mathause:
- [ ] Closes #xxx
 - [ ] Tests added

Adds a helper function to calculate anomalies, which hides not-nice `xr.DataTree` workaround. Turns out this is all nice and dandy until you think about all the ways this could fail. So if an member is missing from the historical simulation it silently removes it from the scenario (inner join). Testing that up-front is annoying (because we don't know the name of the ensemble dimension) so we compare the coords of the anomaly with the original ones.

Also here we see some stuff that is nicer on the new datatree implementation, e.g. a way to iterate over two or more trees (`group_subtrees`). There is some incompatibility in the tests which I don't bother to fix.

### DIFF
--- a/mesmer/core/anomaly.py
+++ b/mesmer/core/anomaly.py
@@ -1,0 +1,75 @@
+import operator
+
+from mesmer.core._datatreecompat import DataTree, map_over_datasets
+
+
+def calc_anomaly(
+    dt: DataTree, reference_period: slice, *, time_dim="time", ref_scenario="historical"
+) -> DataTree:
+    """subtract mean over the reference period
+
+    Parameters
+    ----------
+    dt : DataTree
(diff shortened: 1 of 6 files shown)


mesmer/core/anomaly.py
```diff
@@ -0,0 +1,75 @@
...
+
+        msg = (
+            f"Subtracting the reference changed the coordinates. "
+            f"Most likely because the ref_scenario ({ref_scenario}) is missing "
+            "some ensemble members.\n"
+        )
+
+        raise ValueError(msg)
+
+    return ref
```
mathause, line 75: this is a dummy return as `map_over_datasets` requires it. Will switch to `group_subtrees` when dropping `datatree`

PR #626, opened 2025-04-01
Remove `data` from aod routines, just pass time directly

veni-vidi-vici-dormivi:
This is a mini nit and actually up to personal preference. I felt that passing the temperature data to the aod routines `_predict_volcanic_contributions` and `_load_and_align_strat_aod_obs` is unnecessary and a bit confusing. The data is not actually needed in those, just the time array. Especially in `_predict_volcanic_contributions` the fact that we passed data made me think that we need the smoothed temperature to predict the volcanic contributions but we actually don't. So I rewrote these two routines to just take `time` and `historic_period` as arguments. I understand a bit why it was not like this before because the code might have been a bit prettier before but I think it does make more sense now.

### DIFF
--- a/mesmer/core/volc.py
+++ b/mesmer/core/volc.py
@@ -34,8 +34,9 @@ def _load_and_align_strat_aod_obs(time, hist_period, version="2022"):
     aod = aod.sel(time=hist_period)
 
     # replace time axis of aod -> so they have the same calendar
-    time_hist = time.sel({time.name: hist_period})
-    aod = aod.assign_coords({time.name: time_hist})
+    dim = time.name
+    time_hist = time.sel({dim: hist_period})
+    aod = aod.assign_coords({dim: time_hist})
 
     # expand aod to the full time period
     __, aod = xr.align(time, aod, fill_value=0.0, join="outer")  # type: ignore[assignment]
--- a/mesmer/core/volc.py
+++ b/mesmer/core/volc.py
@@ -3,7 +3,7 @@ import warnings
 import xarray as xr
 
 from mesmer.core._data import load_stratospheric_aerosol_optical_depth_obs
-from mesmer.core.utils import _check_dataarray_form, _assert_annual_data
+from mesmer.core.utils import _assert_annual_data, _check_dataarray_form
 from mesmer.stats import LinearRegression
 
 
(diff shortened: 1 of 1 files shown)


mesmer/core/volc.py
```diff
@@ -3,23 +3,23 @@
...
     load stratospheric aerosol optical depth observations and align them to the to
-    calendar and time of `data` for the `hist_period`
+    calendar and time of `time`.
 
     Parameters
     ----------
-    data : xr.DataArray
-        Data to align aod data to.
+    time: xr.DataArray
+        DataArray containing the time axis to align the aerosol optical depth
```
mathause, line 18:
```suggestion
        DataArray containing the time coords to align the aerosol optical depth
```

mesmer/core/volc.py
```diff
@@ -30,19 +30,12 @@ def _load_and_align_strat_aod_obs(data, hist_period, dim="time", version="2022")
...
-
-    time_hist = time.sel({dim: hist_period})
-
     aod = load_stratospheric_aerosol_optical_depth_obs(version=version, resample=True)
     aod = aod.sel(time=hist_period)
 
     # replace time axis of aod -> so they have the same calendar
-    aod = aod.assign_coords({dim: time_hist})
+    time_hist = time.sel({time.name: hist_period})
+    aod = aod.assign_coords({time.name: time_hist})
```
mathause, line 38:
```suggestion
    dim = time.name
    time_hist = time.sel({dim: hist_period})
    aod = aod.assign_coords({dim: time_hist})
```

mesmer/core/volc.py
```diff
@@ -117,16 +109,36 @@ def fit_volcanic_influence(tas_residuals, hist_period, *, dim="time", version="2
...
-):
+def _predict_volcanic_contribution(time, hist_period, params, version="2022"):
+    """
+    predict volcanic contribution to temperature anomalies using aerosol optical depth
+    observations as proxy
+
+    Parameters
+    ----------
+    time : xr.DataArray
+        DataArray containing the time axis to predict the volcanic contribution for.
```
mathause, line 120:
```suggestion
        DataArray containing the time coords to predict the volcanic contribution for.
```

mathause (commented):
Thanks. Looks good.

mathause (approved):

PR #627, opened 2025-04-03
drop support for xarray-datatree

### LINKED ISSUE #106: internal data structure

### LINKED ISSUE #606: porting `DataTree` to xarray (from datatree)
We wrote the `DataTree` layer with the retired datatree package. Porting this to the new `DataTree` interface in xarray is not-trivial because they added several breaking changes. I have started this endeavor but found several issues right away. I think the only at the moment is pydata/xarray#10013

mathause:
- [x] Closes #606
 - [x] Closed #106
 - [x] Tests added

Drop support for `xarray-datatree` and require `xarray >= 2025.02`. WIP for now. I just saw something I want to address first in a separate PR.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,7 @@ v0.11.0 - unreleased
 
 New Features
 ^^^^^^^^^^^^
-- Implemented new data structure using xarray-datatree, see `Data structure using xarray-datatree`_.
+- Implemented new data structure using ``xr.DataTree``, see `Data structure using DataTree`_.
 - Integrated MESMER-M into the code base, see `Integration of MESMER-M`_.
 - Added number of observations to the output of the AR process (`#395 <https://github.com/MESMER-group/mesmer/pull/395>`_).
   By `Victoria Bauer`_.
@@ -24,8 +24,9 @@ Breaking changes
 - The supported versions of some dependencies were changed
   (`#399 <https://github.com/MESMER-group/mesmer/pull/399>`_,
   `#405 <https://github.com/MESMER-group/mesmer/pull/405>`_,
-  `#503 <https://github.com/MESMER-group/mesmer/pull/503>`_, and
-  `#621 <https://github.com/MESMER-group/mesmer/pull/621>`_):
+  `#503 <https://github.com/MESMER-group/mesmer/pull/503>`_,
+  `#621 <https://github.com/MESMER-group/mesmer/pull/621>`_, and
+  `#627 <https://github.com/MESMER-group/mesmer/pull/627>`_):
 
   ================= ============= =========
   Package           Old           New
@@ -46,7 +47,7 @@ Breaking changes
   **scipy**         not specified 1.12
   **shapely**       not specified 2.0
   **statsmodels**   not specified 0.14
-  **xarray**        2023.04       2024.2
+  **xarray**        2023.04       2025.3
   ================= ============= =========
 
 Deprecations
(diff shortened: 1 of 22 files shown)


ci/requirements/min-all-deps.yml
```diff
@@ -22,8 +22,8 @@ dependencies:
  - scipy=1.12
  - shapely=2.0
  - statsmodels=0.14
- - xarray=2024.2
- - xarray-datatree=0.0.13
+ - xarray=2025.03
+
```
mathause, line 26:
```suggestion
```

mesmer/core/_datatreecompat.py
```diff
@@ -36,11 +23,9 @@ def map_over_datasets(func, *args, kwargs=None):
 else:
     raise ImportError(
         f"xarray version {xr.__version__} not supported - please upgrade to v2025.02 ("
```
mathause, line 25:
```suggestion
        f"xarray version {xr.__version__} not supported - please upgrade to v2025.03 ("
```

mesmer/core/_datatreecompat.py
```diff
@@ -3,19 +3,7 @@
...
-    def map_over_datasets(func, *args, kwargs=None):
-        "compatibility layer for older xarray versions"
-
-        if kwargs is None:
-            kwargs = {}
-
-        return map_over_subtree(func)(*args, **kwargs)
-
-elif Version(xr.__version__) > Version("2025.01"):
+if Version(xr.__version__) > Version("2025.01"):
```
mathause, line 6:
```suggestion
if Version(xr.__version__) >= Version("2025.03"):
```

setup.cfg
```diff
@@ -46,8 +46,7 @@ install_requires =
     scikit-learn >=1.4 # only for the tests
     scipy >=1.12
     statsmodels >=0.14
-    xarray >=2024.2, <2024.10 # upper pin for xarray-datatree
-    xarray-datatree ==0.0.13
+    xarray >=2025.2 # for xr.DataTree
```
mathause, line 49:
```suggestion
    xarray >=2025.3 # for xr.DataTree
```

mathause:
@veni-vidi-vici-dormivi: would you mind reviewing the [changelog](https://mesmer-emulator--627.org.readthedocs.build/en/627/changelog.html#data-structure-using-datatree)? The rest of the PR is actually quite straight-forward. I will have another look at `map_over_datasets` but that probably needs to stay as is for now.

@yquilcaille & @l-pierini this PR drops support for `xarray-datatree` you will need to re-create the environment (or at least upgrade xarray) after this PR is merged

CHANGELOG.rst
```diff
@@ -46,7 +47,7 @@ Breaking changes
   **scipy**         not specified 1.12
   **shapely**       not specified 2.0
   **statsmodels**   not specified 0.14
-  **xarray**        2023.04       2024.2
+  **xarray**        2023.04       2025.3
```
veni-vidi-vici-dormivi, line 50:
```suggestion
  **xarray**        2023.04       2025.03
```

veni-vidi-vici-dormivi (approved):

veni-vidi-vici-dormivi:
Very cool, thanks!

PR #628, opened 2025-04-03
fix pins in environment.yml

mathause:
- [ ] Closes #xxx
 - [ ] Tests added

In #621 I forgot to bump the dependencies in `environment.yml`. Honestly, I am not sure why we even have lower pins (the upper pin of xarray, and the exact pins of regionmask, and datatree are required). No pin means "takes the newest possible version".

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -13,15 +13,15 @@ dependencies:
  - netcdf4
  - numpy
  - packaging
- - pandas>=2.0
+ - pandas>=2.2
  - pooch
  - properscoring
  - pyproj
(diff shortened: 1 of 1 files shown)


mathause:
(For pandas it was converting an upper to a lower pin, but as mentioned these are not needed. 
Still, I am not sure what to do about it so let's continue living in this undecided weird half state that I seem to prefer.)

veni-vidi-vici-dormivi:
> so let's continue living in this undecided weird half state that I seem to prefer.)

Ah yes I am very fond of that one too

PR #635, opened 2025-04-08
remove scripts folder

mathause:
Removes the script folder. Contains only one script with a test that all of our modules can be imported. I don't see how that's necessary...


veni-vidi-vici-dormivi (approved):

mathause:
You were fast! Thanks. The RTD fail must be unrelated.

PR #638, opened 2025-04-08
skip calibration tests on windos

mathause:
- [ ] Closes #xxx
 - [ ] Tests added

Skip tests in test_calibrate_mesmer.py on windows. Win is much slower than linux & mac and these tests check the legacy code base which should be removed soon anyway.


mathause:
ok this takes even longer...

Issue #639, opened 2025-04-10
tighten types


mathause:
We use many union types, e.g.: 

https://github.com/MESMER-group/mesmer/blob/19a03beba5d3a45895834f943b9c2d58cfd498dd/mesmer/core/grid.py#L27-L29

this can be made more precise using `TypeVar`:

```python
from typing import TypeVar, reveal_type

import xarray as xr

T_DataArray_or_Set = TypeVar("T_DataArray_or_Set", xr.Dataset, xr.DataArray)

def f1(data: xr.Dataset | xr.DataArray) -> xr.Dataset | xr.DataArray:
    return data

def f2(data: T_DataArray_or_Set) -> T_DataArray_or_Set:
    return data


reveal_type(f1(xr.Dataset()))
reveal_type(f2(xr.Dataset()))
```

Returns

```
Revealed type is "Union[xarray.core.dataset.Dataset, xarray.core.dataarray.DataArray]"
Revealed type is "xarray.core.dataset.Dataset"
```

also relevant for #632

mathause:
So the `T_DataArraySetTree` is not correct together with `_datatree_wrapper` - while from the outside it's correct (we pass in a `DataTree` and get a `DataTree` back) internally it's different because we are processing a `Dataset`. Not sure what to do about this.

PR #640, opened 2025-04-10
enable functions for xr.DataTree

### LINKED ISSUE #632: enable more functions for DataTree
We have many functions that currently only work for `Dataset` or `DataArray` but should be extended to `DataTree`, e.g.

https://github.com/MESMER-group/mesmer/blob/c93f522291ac102fab256e921bcfc112367ab6f1/mesmer/core/grid.py#L81

We should expand them to accept `DataTree`. Options

1. They may work as is (but see pydata/xarray#10015, so unfortunately many won't)
2. Write a wrapper for each function (as currently done for the stats functions)
3. Use a decorator

```python

import  functools

def datatree_wrapper(func):

    @functools.wraps(func)
    def inner(*args, **kwargs):

        if any(isinstance(arg, xr.DataTree) for arg in args):
            return map_over_datasets(func, *args, kwargs=kwargs)

        return func(*args, **kwargs)

    return inner


@datatree_wrapper
def stack_lat_lon(
    ...
)

```

I think the decorator is the least work/ smallest code change but needs the annoying `isinstance` check. We should still adapt the docstring.

mathause:
- [x] Closes #632
 - [ ] Tests added

Make many functions work with `xr.DataTree`. This will make #521 and #572 a bit simpler. Pretty cool because it's quite trivial. I still need to go over the statistical functions. Unfortunately this does not work for the auto regression functions because the `DataTree` argument (`seed`) is passed as a keyword-only - should we re-order them?

Oh yes, will be annoying to write the tests because it's so many functions. Will probably do minimal test only. The typing will come later (see #639).

https://github.com/MESMER-group/mesmer/blob/19a03beba5d3a45895834f943b9c2d58cfd498dd/mesmer/stats/_auto_regression.py#L377

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -92,6 +92,7 @@ This was originally done with the prototype `xarray-datatree` package. After the
 - Add weighting function for several scenarios (`#567 <https://github.com/MESMER-group/mesmer/pull/567>`_).
 - Add function to compute anomalies over several scenarios stored in a :py:class:`DataTree` (`#625 <https://github.com/MESMER-group/mesmer/pull/625>`_).
 - Add utility functions for :py:class:`DataTree` (`#556 <https://github.com/MESMER-group/mesmer/pull/556>`_).
+- Add a wrapper to allow :py:class:`DataTree` in many data handling functions (`#632 <https://github.com/MESMER-group/mesmer/issues/632>`_).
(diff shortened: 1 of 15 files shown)


mathause:
`any` short-circuits - nice!  (this lowers the cost of the `isinstance` call)

```python
def rt():
    print("called")
    return True

any(f() for f in (rt, rt))
# called
all(f() for f in (rt, rt))
# called
# called
```

mathause:
windows failure is unrelated (I hope this is a one-off)

PR #641, opened 2025-04-11
allow DataTree in `lat_weights` & `weighted_mean`

mathause:
- [ ] Closes #xxx
 - [ ] Tests added

Builds on #640 - need to wait for it before this is interesting to look at.


tests/unit/test_weighted.py
```diff
@@ -95,15 +96,18 @@ def _test_weighted_mean(datatype, **kwargs):
     data = data_lon_lat(datatype, **kwargs)
 
     y_dim = kwargs.get("y_dim", "lat")
-    weights = mesmer.weighted.lat_weights(data[y_dim])
+
+    # TODO: test where we pass DataArray weights
+    # lat = (data["node"].to_dataset() if datatype == "DataTree" else data)[y_dim]
```
mathause, line 101:
```suggestion
```

mesmer/core/weighted.py
```diff
@@ -70,9 +88,21 @@ def weighted_mean(data, weights, dims=None):
 
     """
 
+    # map_over_datasets does not work with kwargs, so need separate funct to be able to pass
+    # weights as kw
```
mathause, line 92: Or should we simplify and just do `def weighted_mean(data, weights, /, *, dims=None):`? (Instead of the separate helper function.)

mathause (commented):
@veni-vidi-vici-dormivi I am a bit unsure about the API changes to allow DataTree in `lat_weights` & `weighted_mean`. Would you mind taking a look?

For reference: 

https://github.com/MESMER-group/mesmer/blob/bbb17c390c9b78f51a2665b3e1d1718d53c21e99/mesmer/core/datatree.py#L202

mesmer/core/weighted.py
veni-vidi-vici-dormivi: Yeah I think that would make more sense

PR #650, opened 2025-04-22
 consolidate package metadata and config

mathause:
- [ ] Closes #xxx
 - [ ] Tests added

This PR consolidates all package meta data and configuration in _pyproject.toml_. This allows to remove _setup.cfg_ and _setup.py_. Also removes the isort and flake8 config (I hope I got used to ruff enough by now... I would have kept it but flake8 does not support _pyproject.toml_). Originally I also intended to remove _MANIFEST.in_, which is where the fun began :slightly_smiling_face: - but would have been an issue independent of this PR.

---

To upload a new release to [pypi](https://pypi.org/project/mesmer-emulator/), two archives are built (1) a wheel and (2) a source distribution (sdist). The files included in these distributions is controlled by the config in _pyproject.toml_ (mostly). The wheel only includes _*.py_ files (unless other files are explicitly added). However, when using _[setuptools_scm](https://setuptools-scm.readthedocs.io/en/stable/)_ (which we do) the sdist includes _all_ files that are tracked by git - including all data files, which currently amounts to more than 100 MB and is over the [file limit](https://pypi.org/help/#file-size-limit) allowed by pypi. Therefore I had to re-add the _MANIFEST.in_ and currently remove the tests folder explicitly.


mathause:
The code formatting changes come from black picking up the supported python version and is equivalent to:

```bash
black --target-version=py310 .
```

PR #653, opened 2025-04-24
make create_equal_dim_names private

mathause:
- [ ] Fully documented, including `CHANGELOG.rst`

Part of #652

PR #654, opened 2025-04-24
remove remove deprecated mask_percentage

### LINKED ISSUE #652: Docs coverage
We could use the [`sphinx coverage extension`](https://www.sphinx-doc.org/en/master/usage/extensions/coverage.html#module-sphinx.ext.coverage) to get a coverage report about which parts of our API are documented. We could even integrate it into CI and block merges when functionality missing documentation. 

However, I spent an afternoon playing around with it and just couldn't get it to run. So if someone else wants to give it a go / has experience with it, please do. For now I put it under "nice to have but not essential". 🥲

mathause:
- [x] Closes #652
 - [ ] Tests added

Actually I realized that we can remove the _regionmaskcompat.py_ file because we could now require regionmask v0.12. 

```
Package           Required             Policy               Status
----------------- -------------------- -------------------- ------
regionmask        0.11    (2023-09-22) 0.12    (2024-03-21) <
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -23,6 +23,8 @@ Breaking changes
   By `Victoria Bauer`_.
 - Removed support for python 3.9 (`#513 <https://github.com/MESMER-group/mesmer/pull/513>`_)
   By `Mathias Hauser`_.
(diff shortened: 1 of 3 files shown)


mathause:
I forgot the changelog in #653 so I added it here

PR #658, opened 2025-04-25
remove unused load_obs functions

mathause:
- [x] Closes #269

These functions are not used. Let's remove them.

PR #660, opened 2025-04-28
RTD: sphinx fail on warning

mathause:
In #658 I forgot to remove the functions from the API docs - this went unnoticed as sphinx does not error on on warnings.


mathause:
This successfully failed :wink: (reading sphinx logs is annoying but this has nothing to do with this change)

mathause:
and now it works :+1:

Issue #666, opened 2025-04-29
re-generate test files


mathause:
Before releasing we should re-generate the output files

1. To fix the rounding of the aod data:
   https://github.com/MESMER-group/mesmer/blob/fc3fe3ab646ee317b8f7e8180fbb602835d0d361/mesmer/core/_data.py#L47-L55
2. Reduce the number of drawn samples in this integration test (its almost 50% of the output data size). Unless there is a good reason to do so many?

   https://github.com/MESMER-group/mesmer/blob/fc3fe3ab646ee317b8f7e8180fbb602835d0d361/tests/integration/test_draw_realisations_mesmer_newcodepath.py#L177

PR #677, opened 2025-05-06
Predictor Arrays in one Dataset per scenario

veni-vidi-vici-dormivi:
Okay, since this is maybe the last change to change something big about the DataTree structure, here goes my idea:

Instead of collecting the different predictor variables (`tas`, `hfds`) in subtrees and those holding scenarios for each variables like this:
```bash
├─ tas
│  ├─ hist
│  ├─ scen1
│  └─ ...
├─ hfds
│  ├─ hist
│  ├─ scen1
│  └─ ...
└─ ...
```

I suggest putting the variables together into datasets and these into a Tree:
```bash
├─ hist
|        datavars: tass, hfds....
├─ scen1
|        datavars: tass, hfds....
└─ ...
```

This is more elegant since the different predictors actually live on the same coordinate system they do fit into a dataset together - so we would be using the data structures in their intended way. We same on memory because we don't have all the coordinated duplicated (which is the rationale behind the intended use).

However, I feel that the Code does get more complicated like this. So I need a second opinion on this.

 - [ ] Tests added


mathause:
Good point - the predictor datatree had to look different than the target which is maybe a bit confusing... How difficult/ annoying is it to add a new variable to the nodes?

---

The third option is to have separate datatrees

```
tas:
├─ hist
├─ scen1
└─ ...

hfds:
├─ hist
├─ scen1
└─ ...
```

veni-vidi-vici-dormivi:
> How difficult/ annoying is it to add a new variable to the nodes?

Not so bad if it shouldn't be generalizable/automated (as in the tests) -- i.e. if you know which ones you want and can do it by names.

The much more annoying part is getting an isomorphic datatree with a subset of variables from one with all variables. Which we need if we eg load a datatree that has both `tas` and `hfds` and somewhere in the code only want one with `tas`.


> This will have an influence on MESMER-X.

Yes.

> The third option is to have separate datatrees

Yes. Also not a big fan of that

mathause:
> > How difficult/ annoying is it to add a new variable to the nodes?
> 
> Not so bad if it shouldn't be generalizable/automated (as in the tests) -- i.e. if you know which ones you want and can do it by names.

> The much more annoying part is getting an isomorphic datatree with a subset of variables from one with all variables. Which we need if we eg load a datatree that has both `tas` and `hfds` and somewhere in the code only want one with `tas`.

Yes this is a general problem - e.g. when we wrote a function for a `DataArray` we cannot easily generalize to `Dataset` or `DataTree`. I am not a fan for passing the variable name, but it may be unavoidable. (Related - there are some functions that check you pass a `Dataset` with only one variable - we'd need to adjust those...)


> > This will have an influence on MESMER-X.
> 
> Yes.
> 
> > The third option is to have separate datatrees
> 
> Yes. Also not a big fan of that

Ok

veni-vidi-vici-dormivi:
Okay I'm done. The new emulations for mesmer are needed because of two things:
(1) for some reasons I used steps = 50 for smoothing hfds and steps = 30 for tas, and now I merged the inputs before smoothing, so both have steps = 30
(2) I moved the calculation of tas**2 to _after_ imposing the volcanic influence bc - duh 
😅

tests/integration/test_calibrate_mesmer_newcodepath.py
```diff
@@ -67,7 +68,7 @@
             True,
             True,
             "tas_tas2_hfds/multi_scen_multi_ens",
-            marks=pytest.mark.slow,
+            # marks=pytest.mark.slow,
```
veni-vidi-vici-dormivi, line 71:
```suggestion
            marks=pytest.mark.slow,
```

mathause:
I have not yet looked at this but if we have to re-generate the test files anyway I'd like to smuggle #666 in beforehand (will probably do this in a separate PR where I'll let the tests fail).

tests/integration/test_calibrate_mesmer_newcodepath.py
```diff
@@ -126,94 +134,60 @@ def test_calibrate_mesmer(
...
+        globmean_smoothed["historical"]["tas"],
         volcanic_params,
     )
 
     # train global variability module
-    tas_resid_novolc = tas_globmean - tas_globmean_smoothed
+    tas_glob_mean = map_over_datasets(lambda ds: ds[["tas"]], globmean)
+    tas_resid_novolc = tas_glob_mean - globmean_smoothed
+    tas_resid_novolc = map_over_datasets(
+        lambda ds: ds.rename({"tas": "tas_resids"}), tas_resid_novolc
```
veni-vidi-vici-dormivi, line 189: there really needs to be an API for this... should we do it or hope `xarray` will do it soon?

tests/integration/test_draw_realisations_mesmer_x.py
```diff
@@ -17,15 +17,15 @@
             "norm(loc=c1 + c2 * __tas__, scale=c3)",
             "expr1",
             False,
-            marks=pytest.mark.slow,
+            # marks=pytest.mark.slow,
```
veni-vidi-vici-dormivi, line 20:
```suggestion
            marks=pytest.mark.slow,
```

tests/integration/test_draw_realisations_mesmer_x.py
```diff
@@ -17,15 +17,15 @@
...
         ),
         pytest.param(
             "ssp585",
             "tasmax",
             "norm(loc=c1 + c2 * __tas__, scale=c3)",
             "expr1_2ndfit",
             False,
-            marks=pytest.mark.slow,
+            # marks=pytest.mark.slow,
```
veni-vidi-vici-dormivi, line 28:
```suggestion
            marks=pytest.mark.slow,
```

tests/unit/test_linear_regression.py
```diff
@@ -254,20 +256,49 @@ def test_lr_predict_exclude_intercept(as_2D, data_type):
...
 
     result = lr.predict(convert_to({}, data_type), exclude={"tas", "intercept"})
-    expected = xr.DataArray([0], dims="x")
+    expected = xr.DataArray([0], dims="x", name="prediction").to_dataset()
     expected = expected if as_2D else expected.squeeze()
 
     xr.testing.assert_equal(result, expected)
 
 
+def test_lr_predict_datatree():
```
veni-vidi-vici-dormivi, line 273: Vici: "If we prove that performing the computation on a `DataTree` works just as performing the computation on the individual `Dataset`s we prove that all the tests that work on `Dataset`s also work on `DataTree`s." let's see 🤞

tests/integration/test_calibrate_mesmer_newcodepath.py
```diff
@@ -222,30 +196,21 @@ def mask_and_stack(ds, threshold_land):
...
+    # train local forced response module
+    # create local gridded data
+    def mask_and_stack(ds, threshold_land):
+        ds = mesmer.mask.mask_ocean_fraction(ds, threshold_land)
+        ds = mesmer.mask.mask_antarctica(ds)
+        ds = mesmer.grid.stack_lat_lon(ds)
+        return ds
 
-        hfds_globmean = mesmer.weighted.global_mean(hfds_anoms)
+    target = map_over_datasets(lambda ds: ds[["tas"]], anoms)
```
mathause, line 207: Opened pydata/xarray#10400

tests/integration/test_calibrate_mesmer_newcodepath.py
```diff
@@ -222,30 +196,21 @@ def mask_and_stack(ds, threshold_land):
...
-    # gridcell can be excluded because it will be mapped in the Linear Regression
-    target = tas_stacked
-    predictors = xr.DataTree.from_dict(
-        {"tas": tas_globmean_smoothed, "tas_resids": tas_resid_novolc}
-    )
     if use_tas2:
-        predictors["tas2"] = tas_globmean_smoothed**2
-    if hfds_globmean_smoothed is not None:
-        predictors["hfds"] = hfds_globmean_smoothed
+        predictors = map_over_datasets(lambda ds: ds.assign(tas2=ds.tas**2), predictors)
```
mathause, line 213: Ah yes, as long as `rename` is not implemented we may as well do this.

tests/integration/test_draw_realisations_mesmer_newcodepath.py
```diff
@@ -8,11 +8,14 @@
 from mesmer.core._datatreecompat import map_over_datasets
 
 
-def create_forcing_data(scenarios, use_hfds, use_tas2):
+def create_forcing_data(scenarios, use_hfds):
```
mathause, line 11:
```suggestion
def load_and_prepare_forcing_data(scenarios, use_hfds):
```

tests/integration/test_draw_realisations_mesmer_newcodepath.py
```diff
@@ -294,44 +273,24 @@ def test_make_realisations(
     )
 
     global_variability = map_over_datasets(
-        xr.Dataset.rename, global_variability, {"samples": "tas"}
+        xr.Dataset.rename, global_variability, {"samples": "tas_resids"}
```
mathause, line 276:
```suggestion
        lambda ds: ds.rename({"samples": "tas_resids"}), global_variability
```

let's align this with the other uses

tests/integration/test_draw_realisations_mesmer_newcodepath.py
```diff
@@ -294,44 +273,24 @@ def test_make_realisations(
...
-    for scen in scenarios:
-        predictors[scen] = xr.DataTree.from_dict(
-            {"tas": tas_forcing[scen], "tas_resids": global_variability[scen]}
-        )
-
-        if tas2 is not None:
-            predictors[scen]["tas2"] = tas2[scen]
-        if hfds is not None:
-            predictors[scen]["hfds"] = hfds[scen]
+    predictors = mesmer.datatree.merge([forcing_data, global_variability])
```
mathause, line 282: Ah here we use it :grin:

mathause (commented):
Thanks - that's pretty neat and a good clean up. Let me know if you don't have time to take care of my suggestions (not that I have time).

tests/integration/test_calibrate_mesmer_newcodepath.py
mathause: Also I am not sure if it actually helps as long as they don't skip empty nodes.

tests/integration/test_draw_realisations_mesmer_newcodepath.py
```diff
@@ -294,44 +273,24 @@ def test_make_realisations(
...
-        local_variability_from_global_var[scen] = xr.DataTree(
-            lr.predict(predictors[scen], exclude=exclude).rename("tas").to_dataset()
-        )
+    # 4.) compute the local variability part driven by global variabilty
+    exclude = {"tas", "intercept"}
+    if use_tas2:
+        exclude.add("tas2")
+    if use_hfds:
+        exclude.add("hfds")
+    local_variability_from_global_var = lr.predict(predictors, exclude=exclude)
```
mathause, line 293: Makes me wonder if we should allow `lr.predict(predictors, only={"tas_resids"})`.

tests/integration/test_draw_realisations_mesmer_newcodepath.py
```diff
@@ -347,13 +306,15 @@ def test_make_realisations(
         buffer=buffer_local_variability,
     )
 
+    # rename datavar in set to be able to add with linear regression predictions
     local_variability = map_over_datasets(
-        xr.Dataset.rename, local_variability, kwargs={"samples": "tas"}
+        xr.Dataset.rename, local_variability, kwargs={"samples": "prediction"}
```
mathause, line 311:
```suggestion
        lambda ds: ds.rename({"samples": "prediction"}), local_variability
```

mesmer/stats/_linear_regression.py
```diff
@@ -21,7 +19,7 @@ def __init__(self):
 
     def fit(
         self,
-        predictors: dict[str, xr.DataArray] | xr.DataTree | xr.Dataset,
```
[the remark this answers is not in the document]
veni-vidi-vici-dormivi: 🫶

tests/integration/test_calibrate_mesmer_newcodepath.py
veni-vidi-vici-dormivi: Let's keep it explicit for now, for the sake of time.

tests/integration/test_calibrate_mesmer_newcodepath.py
```diff
@@ -126,94 +134,60 @@ def test_calibrate_mesmer(
...
-    # create local gridded tas data
-    def mask_and_stack(ds, threshold_land):
-        ds = mesmer.mask.mask_ocean_fraction(ds, threshold_land)
-        ds = mesmer.mask.mask_antarctica(ds)
-        ds = mesmer.grid.stack_lat_lon(ds)
-        return ds
+    anoms = mesmer.anomaly.calc_anomaly(data, REFERENCE_PERIOD)
 
-    tas_stacked = mask_and_stack(tas_anoms, threshold_land=THRESHOLD_LAND)
+    globmean = mesmer.weighted.global_mean(anoms)
```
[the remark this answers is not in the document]
veni-vidi-vici-dormivi: I'm glad I'm not the only one who needs to think twice about these things 😅

tests/integration/test_calibrate_mesmer_newcodepath.py
veni-vidi-vici-dormivi: thanks!

tests/integration/test_draw_realisations_mesmer_newcodepath.py
veni-vidi-vici-dormivi: yes that would be pretty, but not urgent.

mesmer/core/datatree.py
```diff
@@ -297,3 +289,52 @@ def _inner(*args: P.args, **kwargs: P.kwargs) -> T:
         return func(*args, **kwargs)
 
     return _inner
+
+
+from collections.abc import Iterable
+
+from xarray.core import dtypes
+from xarray.core.types import CombineAttrsOptions, CompatOptions, JoinOptions
```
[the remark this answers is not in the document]
veni-vidi-vici-dormivi: this is how pre-commit sorts them 🙈

mathause:
#666 is done (I think) but I need to merge the other 2 PRs first. I'll give this another pass, and we can maybe merge all of them tomorrow.

veni-vidi-vici-dormivi:
> and we can maybe merge all of them tomorrow.

🎉 🎉 

then we'll deal with mesmer-x and the tutorials next

PR #686, opened 2025-05-15
bump dependencies

mathause:
Bump the dependencies once again. I have a _very_ small hope this can help with the failure of `py3.10 ubuntu-latest min-all-deps` in  #645.


mathause:
The difficulty is that I cannot reproduce this failure locally...

PR #687, opened 2025-05-16
add `_assert_required_coords`

mathause:
- [ ] Closes #xxx
 - [ ] Tests added

When stacking scenarios we get `"time"`, `"scenario"`, and `"member"` as non-dimension coordinates, which are not in `obj.dims`, so `"time" in obj.dims` is `False` and thus `required_dims` can error.

As hinted-at in #681


mathause:
Ah yes - this is required for #572

Issue #678, opened 2025-05-08
MESMER-M: calibrate multiple scenarios and ensemble members


mathause:
MESMER-M only used one scenario & ensemble member for calibration (AFAIK). We would like to offer this but need to come up with an approach. #572 attempts this & I just took a closer look, which I want to discuss here. I see three options:

0. Don't offer it
1. Calibrate each ensemble member/ scenario individually and average over them. This follows Lea's approach for the AR process and is currently done in #572
2. Pool all ensemble members and scenarios. This is a bit close to what was done (in a way the historical and ssp585 scenario was pooled), however, there could be discontinuities.

---

I looked at this a bit (for the harmonic model) - see below.

mathause:
We have three sets of params to look at

1. The individual members (im)
2. The average over the im
3. The pooled members (pm)

Technical comment: unfortunately we need to upsample the annualy tas to monthly _before_ pooling (`resample` cannot handle non-monotonic dates).

---

I tried this for the coarse test data - not very precise but I think we can still learn from this.

* **selected order** The selected order of the pm is always equal or larger than the median of the im (except for one out of 5 * 118 = 590 points this also holds for the im). This confused me at first - I expected the pooled one to be in the middle. However, it makes sense: im varies the coeffs for each member, pm cannot do this - so it "stores" variability in additional coeffs.


* **coeffs** We have 4 coeffs per order (https://github.com/MESMER-group/mesmer/pull/512#issuecomment-2324654890):

   $a_0$ | $a_1$ | $b_0$ | $b_1$
     --- | --- | --- | ---
   $\cos \cdot \mathrm{T}$ | $\cos$ | $\sin \cdot \mathrm{T}$ | $\sin$

   Looking at the 3 coefficients of the first order
  

   ![Image](https://github.com/user-attachments/assets/6d056f0b-6c4f-4568-9961-1ad2d2773ce5)

   * The pooled follows more the individual ones than the averaged (e.g. panel a)
   * There seems to be some compensation between the coeff that is multiplied with $\mathrm{T}$ and the one that is not - at least for $\cos$

* **rss** looking at the residual sum of squares I get the following

   what | rss
   --- | --- 
   indiv. | 2.51M
   ave.   | 4.15M
   pooled | 2.59M

   which indicates that the pooled one is much better at capturing the time series than the averaged one

So for me this indicates we should go with the pooled approach. (And raises the question if we should do the same for the auto-regression...)

mathause:
[...]

PR #688, opened 2025-05-16
allow non-dimension time coord in `upsample_yearly_data`

mathause:
- [ ] Closes #xxx
 - [ ] Tests added

Allow non-dimension time coord in `upsample_yearly_data` - so we can pass pooled data in #572 (where `"scenario"` is the dim and `"time"` is a non-dimension coordinate). These 4 lines of code took me almost a day :man_shrugging: :shaking_face: (discovered a new emoji)

### DIFF
--- a/mesmer/core/utils.py
+++ b/mesmer/core/utils.py
@@ -3,6 +3,7 @@ from collections.abc import Callable, Iterable
 from typing import cast
 
 import numpy as np
+import pandas as pd
 import xarray as xr
 
 from mesmer.core.datatree import _datatree_wrapper
@@ -151,8 +152,9 @@ def upsample_yearly_data(
         Upsampled yearly temperature values containing the yearly values for every month
         of the corresponding year.
     """
-    _assert_required_dims(yearly_data, "yearly_data", required_dims=time_dim)
-    _assert_required_dims(monthly_time, "monthly_time", required_dims=time_dim)
+
+    _assert_required_coords(yearly_data, "yearly_data", required_coords=time_dim)
+    _assert_required_coords(monthly_time, "monthly_time", required_coords=time_dim)
 
     # read out time coords - this also works if it's already time coords
     monthly_time = monthly_time[time_dim]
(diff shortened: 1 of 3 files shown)


mesmer/core/utils.py
```diff
@@ -151,8 +152,9 @@ def upsample_yearly_data(
         Upsampled yearly temperature values containing the yearly values for every month
         of the corresponding year.
     """
-    _assert_required_dims(yearly_data, "yearly_data", required_dims=time_dim)
-    _assert_required_dims(monthly_time, "monthly_time", required_dims=time_dim)
+
+    _assert_required_coords(yearly_data, "yearly_data", required_coords=time_dim)
```
mathause, line 156: @veni-vidi-vici-dormivi do you have an opinion if we should rename `time_dim` to `time_coord`?

(For me `time_coord` sounds much more like the coordinates themselves than the it's name (in contrast to `time_dim`) and `time_coord_name` is too long IMHO. But I am happy to get your opinion.)

mesmer/core/utils.py
```diff
@@ -163,15 +165,27 @@ def upsample_yearly_data(
             "Length of monthly time not equal to 12 times the length of yearly data."
         )
 
-    # make sure monthly and yearly data both start at the beginning of the period
-    year = yearly_data.resample({time_dim: "YS"}).bfill()
-    month = monthly_time.resample({time_dim: "MS"}).bfill()
+    # we need to pass the dim (`time_dim` may be a no-dim-coordinate)
+    # i.e., time_dim and sample_dim may or may not be the same
+    (sample_dim,) = monthly_time.dims
```
mathause, line 170: Took me too long to realize we can read out the `sample_dim` from the coords. Like this we don't need to pass them, which would have been awkward.

mathause:
I think with this I finally made some progress for a nice API towards #678

mesmer/core/utils.py
mathause: @veni-vidi-vici-dormivi do you have an opinion on this?

veni-vidi-vici-dormivi: Sorry for missing this. Uhm I guess as long as it should be a string I would opt for `time_dim`. We also use `time_dim` in `predict_harmonic_model()` and `fit_harmonic_model()`. But then again we could rename them all to `dim`, which is what use in `fit_auto_regression()` (but not in `draw_auto_regression_*`). `time_coord` also for me would suggest I should pass a `xr.DataArray`.

mathause: No worries - thanks for the feedback

PR #699, opened 2025-06-06
map_over_datasets: allow Dataset in all positions

mathause:
- [ ] Closes #xxx
 - [ ] Tests added

Extracted from #677 - so this is actually by @veni-vidi-vici-dormivi - thanks! The only think I changed is 

```diff
-if any((not ds.coords and not ds.data_vars and not ds.dims) for ds in ds_args):
+if any((not ds.coords and not ds.data_vars) for ds in ds_args):
```

I think a ds can only have `dims` if it either has `coords` or `data_vars`. I'll merge fast.


veni-vidi-vici-dormivi:
Thank you!! 🙏

PR #709, opened 2025-06-16
draw_auto_regression_monthly: allow cov None

mathause:
There is a line which suggests that we can pass `covariance=None` to `draw_auto_regression_monthly`

https://github.com/MESMER-group/mesmer/blob/04311c895570205b5d81f4f2877ca2b66e1c8eef/mesmer/stats/_auto_regression.py#L1153

However, some more changes are required to make this work. Alternatively we could just remove the possibility.

@veni-vidi-vici-dormivi? (very low priority)


veni-vidi-vici-dormivi:
That doesn't make a lot of sense, does it? Remove it...

PR #711, opened 2025-06-16
GHA: add more environment info

mathause:
- [ ] Closes #xxx
 - [ ] Tests added

I am particularly interested to find out how many cores and threads are used. However, I am not sure if we can get this info on the virtual machines and on windows.


mathause:
(so slightly brittle as psutils is an indirect dependency)

Issue #349, opened 2023-12-06
restrict number of used numpy threads


mathause:
As already mentioned by @jschwaab in #168, estimating the localization radius is resource heavy. We should restrict the number of threads used for this. 

https://stackoverflow.com/q/30791550/3010700

mathause:
```python
import threadpoolctl
threadpoolctl.threadpool_limits(limits=12)

# or just for one function:

with threadpoolctl.threadpool_limits(limits=12):
    np.linalg.eig(arr)


threadpoolctl.threadpool_info()

```

Interestingly I got better performance by restricting the number of threads to the number of physical cpus.

PR #712, opened 2025-06-16
TEST: limit threads

mathause:
Test towards #349

I just tested limiting the number of threads locally and locally the difference to estimate the local covariance is massive. For #572 it's 2 min vs 4 s (!). So let's try that out on GHA. 

#349


mathause:
After I learned that windows uses 2 threads per default I want to try if it is slower when we run it with more.

PR #713, opened 2025-06-19
restrict threads and make them selectable

### LINKED ISSUE #349: restrict number of used numpy threads

mathause:
- [x] Closes #349, supersedes and therefore closes #712
 - [x] Tests added

My tests in #712 were a bit underwhelming, however, I think this is still worth it.

- Large calibrations are likely run on linux systems & I saw good gains on our systems
- I get considerable gains on my old laptop
- There seems to be no speed loss on Mac/ Windows

I think it's more likely to run several jobs in parallel on large systems, so I made an upper limit. The `16` were semi-randomly chosen (`2 ** 4`), happy to adjust to something else.

I think there are currently only two places where we do a matrix decomposition.

I added `options`, a bit overkill for one option :man_shrugging:. I copied it from regionmask (originally adjusted from xarray).

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -73,6 +73,9 @@ jobs:
         echo ""
         echo "Number of cpus and threads"
         python -c "import psutil; print(f'cpus: {psutil.cpu_count(logical=False)}; threads: {psutil.cpu_count()}')"
+        echo ""
+        echo "threadpool info:"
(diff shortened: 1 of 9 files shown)


mathause:
No idea why the windows test failed. I cannot reproduce locally...

mathause:
windows only uses 2 threads on windows on GHA - interesting.

PR #715, opened 2025-06-19
rename stack -> pool

mathause:
1. In #572 I use `_stack_datatree` in the example so this should be made public.
2. we have `mask_and_stack`, so I ended up with `tas_stacked_stacked_y` :grimacing: so my suggestion is to rename it to `pool`

(I am also not sure `datatree` is the best namespace but #348 :disappointed:)

@veni-vidi-vici-dormivi do you have thoughts on this?

### DIFF
--- a/mesmer/core/datatree.py
+++ b/mesmer/core/datatree.py
@@ -131,7 +131,7 @@ def collapse_datatree_into_dataset(
     return ds
 
 
-def _stack_datatree(
+def pool_scenarios(
     dt: xr.DataTree,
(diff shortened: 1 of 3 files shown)


mathause:
Here we use `broadcast_and_pool_scenarios` for the auto regression we have `select_ar_order_scen_ens` - so should these be `broadcast_and_pool_scen_ens`?

mathause:
(ah yes will need to update mesmer-x again...)

PR #716, opened 2025-06-20
turn YeoJohnson into a class

mathause:
- [ ] towards #517
 - [ ] Tests added

Not sure if the approach with the subclass is a good API, and whether we should do this at all. And sorry anyway for all these last-minute changes :unamused:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -224,7 +224,8 @@ Yeo-Johnson power transformer
    `#475 <https://github.com/MESMER-group/mesmer/pull/475>`_, and
    `#425 <https://github.com/MESMER-group/mesmer/pull/425>`_)
 - add tests (`#430 <https://github.com/MESMER-group/mesmer/pull/430>`_)
-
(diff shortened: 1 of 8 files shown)


veni-vidi-vici-dormivi:
Hm sorry for being late on this but why not 

```python
class PowerTransformer:
   def __init__(self, method):
        if method == "logistic":
            self.lambda_function = logistic_lambda_func
 ```
?

mathause:
> Hm sorry for being late on this but why not
> 
> ```python
> class PowerTransformer:
>    def __init__(self, method):
>         if method == "logistic":
>             self.lambda_function = logistic_lambda_func
> ```
> 
> ?

Not late at all. Yes I like that - this simplifies the code and the docs.

mathause:
> > Yes I like that - this simplifies the code and the docs.
> 
> That dopamine will last me a few months 😄❤️

:hugs:

mathause:
Ok in it goes :crossed_fingers:

PR #718, opened 2025-06-23
YeoJohnson: add constant lambda function

### LINKED ISSUE #517: allow different lambda functions (besides logistic)?

mathause:
- [ ] Closes #517
 - [ ] Tests added

After #716 it is now possible to add other covariate functions. I may use this for the example.

### DIFF
--- a/tests/unit/test_power_transformer.py
+++ b/tests/unit/test_power_transformer.py
@@ -37,6 +37,11 @@ def test_yj_transformer_bounds_first_guess():
     first_guess = np.array([1.0, 0.0])
     np.testing.assert_equal(yj_transformer.first_guess, first_guess)
 
+    yj_transformer = YeoJohnsonTransformer("constant")
+    np.testing.assert_equal(yj_transformer.bounds, None)
+    first_guess = np.array([1.0])
+    np.testing.assert_equal(yj_transformer.first_guess, first_guess)
+
 
 @pytest.mark.parametrize(
     "coeffs, t, expected",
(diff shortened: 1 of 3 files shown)


tests/unit/test_power_transformer.py
```diff
@@ -208,15 +230,15 @@ def test_yeo_johnson_optimize_lambda_sklearn():
...
+    yj_transformer = YeoJohnsonTransformer(name)
     ourfit = yj_transformer._optimize_lambda_np(local_monthly_residuals, yearly_T)
     result = yj_transformer.lambda_function(ourfit, yearly_T_value)
     sklearnfit = PowerTransformer(method="yeo-johnson", standardize=False).fit(
         local_monthly_residuals.reshape(-1, 1), yearly_T.reshape(-1, 1)
     )
     expected = sklearnfit.lambdas_
 
-    np.testing.assert_allclose(np.array([result]), expected, atol=1e-5)
+    np.testing.assert_allclose(np.array([result]), expected, atol=1e-4)
```
mathause, line 241: This seems a bit strange - the simpler covariate function has the worse precision. However, this could be fixed by a smaller `tol` in `minimize` (which would, however, make the whole thing slower).

PR #733, opened 2025-07-03
drop python 3.10 support

mathause:
- [ ] Closes #xxx
 - [ ] Tests added


mathause:
As per https://github.com/MESMER-group/mesmer/pull/732#issuecomment-3031308243 and yes our minimum version policy allows it

PR #739, opened 2025-07-07
drop scipy pin

### LINKED ISSUE #728: unpin scipy
#727

mathause:
- [ ] Closes #728
 - [ ] Tests added

statsmodels released a new version so we can drop the scipy pin (may still take a while until it's on conda) and yes there are now combinations that do not work


mathause:
(scipy was pinned in the https://github.com/conda-forge/statsmodels-feedstock - so can remove the pin now - and it will do the right thing)

PR #753, opened 2025-07-22
rename core to _core

mathause:
- [ ] Closes #xxx
 - [ ] Tests added

Not entirely sure about this (mostly because of the way our docs are rendered). But would make core explicitly private.


mathause:
I do prefer this - the only issue is the way the docs are rendered

image

Which would make #348 more important to solve 😐

mathause:
I committed a file that should not have been.

mathause:
I open a new PR after #773 and #771

PR #759, opened 2025-07-23
make more args kw only

### LINKED ISSUE #245: make arguments keyword only
Make a number of arguments keyword only - this enforces good practice and can help deprecating arguments.

mathause:
- [ ] Closes #245
 - [ ] Tests added

Went over all public functions in _stats_, _mesmer_x_, and _core_. It's not entirely consistent if `dim` is kw only or not.


mathause:
We now really see the benefits of running the notebooks every time.

PR #767, opened 2025-07-25
MESMER-M: test calibration of multiple scen & ens

### LINKED ISSUE #678: MESMER-M: calibrate multiple scenarios and ensemble members

### LINKED ISSUE #719: MESMER-M: multi-ens-multi-scen leftovers
There are two open points from https://github.com/MESMER-group/mesmer/pull/572#issuecomment-2988659770

- [ ] integration test for multi-ens-multi-scen MESMER-M
- [x] changelog entry for all multi-ens-multi-scen MESMER-M related changes

I will address the open points in a separate PR. Let's see maybe with a `"constant"` Yeo-Yohnson transformer this is a bit faster.

mathause:
- [ ] Closes #719
 - [ ] Closes #678
 - [ ] Tests added

The second test case has no result files yet - so it still fails. Also need to update the name of the output files.

Interesting how using a constant transformation makes the test faster even with more samples...

### DIFF
--- a/tests/integration/test_calibrate_mesmer_m.py
+++ b/tests/integration/test_calibrate_mesmer_m.py
@@ -77,8 +77,38 @@ def load_data(filecontainer):
     return out
 
 
-@pytest.mark.slow
-def test_calibrate_mesmer_m(test_data_root_dir, update_expected_files):
+@pytest.mark.parametrize(
+    (
+        "scenario",
+        "n_ens",
+        "yj_transformation",
+        "outname",
+    ),
+    [
+        pytest.param(
+            ["ssp585"],
+            "one",
+            "logistic",
+            "tasmax/one_scen_one_ens",
+            marks=pytest.mark.slow,
+        ),
+        pytest.param(
+            ["ssp126", "ssp585"],
+            "many",
+            "constant",
+            "tasmax/one_scen_one_ens",
+            marks=pytest.mark.slow,
+        ),
+    ],
+)
+def test_calibrate_mesmer_m(
+    scenario,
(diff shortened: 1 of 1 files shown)


veni-vidi-vici-dormivi:
Saving the monthly time doesn't work anymore when we have stacked samples:
`m_time = tas_stacked_m.time.rename("monthly_time")`

We should switch to saving the time of `tas_m_orig`: `m_time = xr.concat([tas_m_orig["historical"].time, tas_m_orig[scenario[0].time], dim="time")`.  I'll let you decide if you want to do it with and if else statement or come up with a way to generalize it.

mathause:
In the tests I am loading the data via `datatree` for both cases - so I decided to just get the time from the original monthly tas datatree:

```python
    time_hist = tas_m_orig["historical"].ds[["time"]]
    time_proj = tas_m_orig[scenario[0]].ds[["time"]]

    m_time = xr.concat([time_hist, time_proj], dim="time")
    m_time = m_time.time.rename("monthly_time")
```

Issue #769, opened 2025-07-29
Running out of memory fast when drawing MESMER-M emulations


veni-vidi-vici-dormivi:
Drawing emulations for MESMER-M is (at least?) 12 times more memory intensive than for MESMER. I cannot draw much more than 100 emulations on the MacBook Pro with 18 GB of RAM (the process gets killed or runs into a segmentation fault at about 60 GB of swap usage).
Something related might be #165 (2).

veni-vidi-vici-dormivi:
https://github.com/MESMER-group/mesmer/blob/be8455e14ff6ea16447f336a9c981c66656e0d59/mesmer/stats/_power_transformer.py#L44

Not optimal either, even though it happens in an `apply_ufunc` along `sample_dim` so I think it deletes the doubled array after every gridcell and realisation.

mathause:
This is the required RAM to create the emulations or the size of the emulations? The second case would not be possible to store in RAM no matter what we optimize.

veni-vidi-vici-dormivi:
No that's the total size of the emulations.

PR #772, opened 2025-07-29
Reference copy `innovations` into `out` for `draw_auto_regression_monthly_np`

veni-vidi-vici-dormivi:
related to #769


mathause (approved):

veni-vidi-vici-dormivi:
Jup, that halved the time needed to draw emulations with mesmer-m 😅

PR #773, opened 2025-07-30
move data modules to mesmer namespace

### LINKED ISSUE #348: mesmer module naming and structure

mathause:
- [ ] Closes #348
 - [ ] Tests added

Move data processing modules directly to the mesmer namespace. I am not happy with this but don't know any better, either :slightly_frowning_face:


mathause:
Will probably move `mesmer.core.utils.upsample_yearly_data` to `mesmer.resample`. Not sure about `mesmer.data`.

PR #788, opened 2025-08-28
test without integration & legacy tests (do not merge)

mathause:
- [ ] Closes #xxx
 - [ ] Tests added


mathause:
This is pretty interesting and could be a good exercise to fix (if we had time)...

- there seem to be no unit test for `ProbabilityIntegralTransform`
- Maybe [test_volc.py](https://github.com/MESMER-group/mesmer/blob/4e4fca9569273faec69d7c5d354b4678d42d8e11/tests/integration/test_volc.py) should be a unit test? (instead of integration)

PR #832, opened 2025-12-17
avoiding degradation of fit in _fit_np

yquilcaille:
While training mrso, i noticed issues during the training. Basically, the Cholesky decomposition failed for all localization radii, because the matrix was symetric, but not all eigenvalues were positive. When digging, i discovered that the training of conditional distributions was failing for many grid points. It was caused by two issues. 

1. I noticed that the minimization during `ConditionalDistribution._fit_np` was for some reason finishing on a `m.x` with a higher function value than the first guess... Which should not happen. I fixed that checking at the end that if `m.x` is returned only if its minimization value is actually below than the value corresponding to the first guess.
2. Then i noticed in the 2nd round of training infinite values. It indicates that the first guess of this step, ie the result of the 1st round smoothed over space, was sometimes leading to values out of boundaries. Thus the failed fit should not use the first guest provided during minimization, which is smoothed, but the values before smoothing. Thus i introduced a fail safe copy. It ensures that if the 2nd round of training fails because of the smoothing, then we are not degrading the performance of the fit, but instead just come back to the level of the 1st round.

I'm not adding any test here, i dont think that there is a need for this step. I will check whether it crashes those existing...


mathause (commented):
How about renaming `fg_failsafe` to `fg_initial` (or `fg_original`)?

---

I fear that we have non-sensible targets or wrong assumptions if the minimization actively degrades the estimate. Do you know in which step this happens?

But yes minimize behaves strangely sometimes.

mathause:
Would be interesting to have a test case to play around with & to understand what's going on. But I guess that's difficult to extract. Could it be the `threshold_min_proba`?

yquilcaille:
My best bet is that during the first guess, step 5 is based on `self._fg_fun_nll_no_checks`, then if the solution doesnt pass tests (eg `threshold_min_proba`), the step 6 is to use `self._fg_fun_nll_cubed` an attempt to improve the checks. This solution is then used in the minimization with NLL and all tests. It may be that the solution of the cubing NLL works in terms of tests but isnt optimal for the usual NLL. How to solve that? Unsure. Digging into the step 6 with the inclusion of tests is really technical.

I would say that we have here is a relatively good version, although not perfect.

mathause (commented):
The upstream issue was resolved, so that should no longer be an issue.

PR #837, opened 2026-01-09
MESMER-MTP: precipitation

sarasita:
- [ ] Closes #xxx
 - [ ] Tests added


mesmer/stats/_regularized_glm.py
```diff
@@ -0,0 +1,182 @@
...
+        closest_locations:
+            DataArray with dims (gridcell, closest_gridcells)
+        """
+
+        gridcells = tas.gridcell.values
+        months = tas.month.values
+
+        n_years = tas.sizes["year"]
+        n_closest = closest_locations.sizes["closest_gridcells"]
+        n_cov = 1 + 2 * n_closest
```
mathause: I have not looked at this in details but this is very use case specific - check if we could move this out of the function somehow.

mesmer/stats/_xarray_pipelines.py
```diff
@@ -0,0 +1,283 @@
...
+                coords = {
+                    self.feature_dim: self.feature_coords,
+                    "component": np.arange(param.shape[0]),
+                }
+            return xr.DataArray(param, dims=dims, coords=coords)
+        else:
+            return param
+
+
+# class XarrayPipeline:
```
mathause: can we delete this commented block?

mathause (commented):
You were not in the office therefore I ask here:

- What is the difference between the _example_mesmer_mtp_ and _example_mesmer_mtp_precipitation_ notebooks? I see that the first one does not load pr, but doesn't the "p" in "mtp" stand for precipitation?

- There seems to be two ways to calculate a PCA:
  1. the `fit_principal_components` function (which applies a `StandardScaler` and then the `PCA`); used in the _example_mesmer_mtp_ notebook.
  2. the `XarrayPipeline`, where you explicitely pass a `StandardScaler` and `PCA` as `steps`; used in the _example_mesmer_mtp_precipitation_ notebook

   Are these two ways equal? Do we need to keep both? Or which one can we remove?

mesmer/stats/_regularized_glm.py
```diff
@@ -0,0 +1,287 @@
...
+class FeaturewiseRuleGLM:
+    """
+    Feature-wise GLM with dynamic design rule for X per feature.
+    Supports multiple feature dimensions or none.
+    """
+
+    def __init__(
+        self,
+        design_rule,
+        alpha_grid,
```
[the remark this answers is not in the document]
sarasita: Mh... I suppose either way would be fine.

mathause:
scikit-learn does also seem to loop the all class attributes and collects the ones that end in `_` and do not start with an underscore:

https://github.com/scikit-learn/scikit-learn/blob/f3182980c56943664f6b1ff30e8d0c436bd7139e/sklearn/base.py#L361

PR #864, opened 2026-04-13
YeoJohnson: inverse_transform warn invalid

mathause:
- [x] Closes #859
 - [x] Tests added


mathause:
@yquilcaille/ @sarasita: Do you create emulations for MESMER-M for 'real' data? Could you run your emulations again and test if you get warnings for `yj_transformer.inverse_transform`?

See also https://github.com/MESMER-group/mesmer/issues/859#issuecomment-4236290052 for the range of combinations of lambda and data which yield valid back-transformations?

PR #881, opened 2026-05-08
`fit_harmonic_model`: return residuals separately

### LINKED ISSUE #874: adding residuals to fit results was a bad idea
We add `residuals` to some of the results. This was a very bad idea because if we are not careful these get saved as parameters - which is of course exactly what happened. For the uploaded [mesmer_m](https://libdrive.ethz.ch/index.php/s/o5MBHatA3GTw53E?path=%2Fparameters_mesmer_m_v1) params, for e.g. _CanESM5_ssp119-ssp126-ssp245-ssp370-ssp460-ssp585-ssp534-over_ is now **18G** instead of **1.7G** a more than 10 fold increase.

So instead of ~ 20 GB of MESMER-M params we ended up with ~ 200 GB.

In the tests we do remove them:

https://github.com/MESMER-group/mesmer/blob/a302032887974609633f564514f121df7a35cea4/tests/integration/test_calibrate_mesmer_m.py#L251-L252

To illustrate this is are the harmonic model params:

```
<xarray.Dataset> Size: 9GB
Data variables:
    selected_order  (gridcell) int64 24kB ...
    coeffs          (gridcell, coeff) float64 568kB ...
    residuals       (sample, gridcell) float64 9GB ...
```

---

(We could have saved another 800 MB of the remaining 1700 MB if we only saved `localized_covariance` and not the `covariance`.)

Blrg

cc @yquilcaille @veni-vidi-vici-dormivi

mathause:
- [x] Closes #874, companion to #880
 - [x] Tests added

Again, we can still decide not to do this. Contains changelog entry for #880 as well.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,6 +19,10 @@ New Features
 Breaking changes
 ^^^^^^^^^^^^^^^^
 
+- :py:func:`fit_harmonic_model` and :py:func:`fit_auto_regression_monthly` not return the residuals as
+  separate :py:class:`xr.DataArray` instead of merging it into the :py:class:`xr.Dataset` containing the
+  fitted parameters   (`#880 <https://github.com/MESMER-group/mesmer/pull/880>`_, and `#881 <https://github.com/MESMER-group/mesmer/pull/881>`_).
+  By `Mathias Hauser`_.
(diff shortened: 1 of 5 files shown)


mathause:
Close in favor of #882

PR #882, opened 2026-05-11
HarmonicModel class

mathause:
- [ ] Closes #xxx
 - [ ] Tests added

Alternative to #881 - add a `HarmonicModel` class. There are elements that are nicer and some that are less nice than the function based approach. I am not convinced if `residuals` should really be a `method` :-|

I have not yet removed the duplicated residuals computation. Also changelog entry is missing.


veni-vidi-vici-dormivi:
Puh idk... [`sklearn.LinearRegression`](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LinearRegression.html) has neither for example... having `.residuals()` as a method has the slight advantage that it makes sense then that `to_netcdf()` does not save the residuals...

mathause:
Yes I also thought if we should just remove it. But due to DataTree shenanigans we cannot remove it from `LinearRegression`. We could remove it here, though...

veni-vidi-vici-dormivi:
Please still do the changelog entry @mathause 🙏

mathause:
> Please still do the changelog entry @mathause 🙏

Thanks! Done in #890

veni-vidi-vici-dormivi:
Thanks! Ah and the duplication thing is fine?

PR #886, opened 2026-05-12
drop python 3.11 support

mathause:
Let's see it that speeds up solving the environment.


mathause:
It does not but let's merge anyway...

PR #888, opened 2026-05-12
fix typos

mathause:
```bash
codespell --skip=".git,devel,build" --ignore-words-list "varn, THIRDPARTY, strat, nD" . --write-changes -i 3
typos
typos -w
```

Again a nice list of typos has accumulated.


veni-vidi-vici-dormivi (approved):
initial seems to be a hard word 😂

PR #893, opened 2026-05-12
Deduplicate calculation of residuals in HarmonicModel

veni-vidi-vici-dormivi:
The doctoring in `HarmonicModel.fit()` said it returns the residuals, but it didn't. This bothered me too much so I cleaned it up. Now the residuals are only calculated in `HarmonicModel.residuals()` and there is also no private property. Following #881

### DIFF
--- a/mesmer/stats/_harmonic_model.py
+++ b/mesmer/stats/_harmonic_model.py
@@ -168,7 +168,7 @@ def _fit_fourier_coeffs_np(yearly_predictor, monthly_target, first_guess):
     coeffs : array-like of shape (4*order)
         Fitted coefficients of Fourier series.
 
-    mse : Float
+    mse : float
         Mean-squared error.
 
     """
--- a/mesmer/stats/_harmonic_model.py
+++ b/mesmer/stats/_harmonic_model.py
@@ -168,8 +168,8 @@ def _fit_fourier_coeffs_np(yearly_predictor, monthly_target, first_guess):
     coeffs : array-like of shape (4*order)
         Fitted coefficients of Fourier series.
 
-    preds : array-like of shape (n_years*12,)
-        Predicted monthly values.
+    mse : Float
+        Mean-squared error.
 
     """
 
@@ -242,8 +242,6 @@ def _fit_fourier_order_np(yearly_predictor, monthly_target, max_order):
         Selected order of Fourier Series.
     coeffs : array-like of size (4 * max_order,)
         Fitted coefficients for the selected order of Fourier Series.
-    predictions : array-like of size (n_years*12,)
-        Predicted monthly values from final model.
     """
 
     current_min_score = float("inf")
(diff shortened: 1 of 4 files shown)


veni-vidi-vici-dormivi:
Okay, I'm confused @mathause, I thought you said tests would fail if we do this. Did I miss something?

mesmer/stats/_harmonic_model.py
```diff
@@ -372,14 +366,12 @@ def _fit_harmonic_model(
 
     coeffs = coeffs.assign_coords({"coeff": np.arange(coeffs.sizes["coeff"])})
 
-    resids = monthly_target - (yearly_predictor + preds)
```
mathause, line 375: You can now remove the computation of `preds`, here and in `_fit_fourier_order_np` as they are no longer used. But tested in

https://github.com/MESMER-group/mesmer/blob/e09c52e09784e17453d5276b59752cc104006f12/tests/unit/test_harmonic_model.py#L92

mesmer/stats/_harmonic_model.py
```diff
@@ -449,8 +434,6 @@ def fit(
 
         self._params = params
 
-        self._residuals = residuals
```
mathause, line 452: I thought I removed this 😑

mathause (commented):
Thanks! It seems I did not remove all of the code in the class.

veni-vidi-vici-dormivi:
I'm sorry, I *really* didn't think you would be up. I found it myself 🙈

mathause:
Apparently mosquito season started :slightly_frowning_face: and I wake up when they bite me 🦟

mathause:
And why were you still up?

mesmer/stats/_harmonic_model.py
```diff
@@ -168,8 +168,8 @@ def _fit_fourier_coeffs_np(yearly_predictor, monthly_target, first_guess):
     coeffs : array-like of shape (4*order)
         Fitted coefficients of Fourier series.
 
-    preds : array-like of shape (n_years*12,)
-        Predicted monthly values.
+    mse : Float
```
mathause:
```suggestion
    mse : float
```

mesmer/stats/_harmonic_model.py
```diff
@@ -359,30 +347,29 @@ def _fit_harmonic_model(
...
         "selected_order": selected_order,
         "coeffs": coeffs,
     }
 
-    return xr.Dataset(data_vars), resids.transpose(sample_dim, ...)
+    return xr.Dataset(data_vars)
 
 
 class HarmonicModel:
+    """HarmonicModel class to fit a Fourier Series to monthly data using yearly predictors."""
```
mathause: Can you move that to `__init__`?

mathause (approved):
Thanks yes this is better!

mesmer/stats/_harmonic_model.py
veni-vidi-vici-dormivi: I can but in Linear regression we also have it under the class and not in the init... Ok but for the PowerTransformer it's also in the init. I moved the ones from the HarmonicModel and the LinearRegression to the inits now.

veni-vidi-vici-dormivi:
> And why were you still up?

Complicated... Our train is leaving at 11:30 and I want to get this done, but also Paul is building a NAS and was up debugging it as well, otherwise I would probably just have gone to bed 😅

PR #911, opened 2026-07-13
upstream dev CI script: strict mode

mathause:
The upstream dev CI should fail earlier if e.g., a package is not available in the nightlies.


mathause:
The CI failed because scikit-learn was missing - the error is much clearer now.

PR #912, opened 2026-07-13
draw_auto_regression_correlated: name of covariance

mathause:
- [ ] Closes #xxx
 - [ ] Tests added

@yquilcaille: requires the `covariance` in `draw_auto_regression_correlated` to be named _localized_covariance_adjusted_. Not nice but should help to avoid the problem of passing the wrong covariance. I think this is independent from what we do with the wrong uploaded data.

If you are ok with this change I suggest to release version 1.1.0 in the next days.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,7 +10,10 @@ New Features
 Breaking changes
 ^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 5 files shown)


mathause:
I thought about using a warning only - but experience shows that warnings are ignored...

